PR #2, opened 2020-11-27
Make async stuff work reliably

tom-pytel:
This is not done yet but I am opening the PR before the changes get too big so you can see the direction I am going and comment. I am implementing basically the same mechanism as in the Python agent to allow async stuff to work without losing spans or weird parent-child relationships where they should be siblings. Here are a few notes / questions so far:

* Context.capture() and restore() will be made redundant / unnecessary by this?

* ContextManagere.withSpan() export to user?

* Node 10 does not have AsyncLocalStorage, will need to provide custom substitute class for this (using async_hooks.createHook init and destroy to track parent-child relationships).

* I had this question with the Python agent whether entry spans should always start with a new context but decided against it because maybe someone wants a previous local context for some debug or other reason to chain before the entry context? So I did the same here and disabled fresh context on Entry span but rather just let it use the normal inheritance chain, which should eventually give a new context unless the user specifically prevents it.

* Has problem as Python agent with background updater not firing if application exits very quickly, should not exit until all data sent, probably should hook 'beforeExit' event and flush segments, should do same in Python.

### DIFF
--- a/src/trace/context/Context.ts
+++ b/src/trace/context/Context.ts
@@ -24,7 +24,6 @@ import { ContextCarrier } from './ContextCarrier';
 
 export default interface Context {
   segment: Segment;
-  spans: Span[];
 
   newLocalSpan(operation: string): Span;
 
@@ -36,13 +35,13 @@ export default interface Context {
 
   stop(span: Span): boolean;
 
+  async(span: Span): void;
+
+  resync(span: Span): void;
+
   currentSpan(): Span | undefined;
 
   capture(): Snapshot;
 
   restore(snapshot: Snapshot): void;
-
-  async(span: Span): void;
-
-  await(span: Span): void;
 }
--- a/src/trace/context/Context.ts
+++ b/src/trace/context/Context.ts
@@ -35,8 +35,13 @@ export default interface Context {
 
   stop(span: Span): boolean;
 
+  /* This should be called just before a span is passed to a different async context, like for example a callback from
+     an asynchronous operation the code belonging to the span initiated. After this is called a span should only call
+     .resync() or .stop(). See HttpPlugin.interceptClientRequest() in plugins/HttpPlugin.ts for example of usage. */
   async(span: Span): void;
 
+  /* This should be called upon entering the new async context for a span that has previously executed .async(), it
+     should be the first thing the callback function belonging to the span does. */
   resync(span: Span): void;
 
   currentSpan(): Span | undefined;
--- a/src/trace/context/ContextManager.ts
+++ b/src/trace/context/ContextManager.ts
@@ -18,18 +18,76 @@
  */
 
 import Context from '../../trace/context/Context';
+import Span from '../../trace/span/Span';
 import SpanContext from '../../trace/context/SpanContext';
 import { AsyncLocalStorage } from 'async_hooks';
 
-const store = new AsyncLocalStorage<Context>();
+type AsyncState = {context: Context, spans: Span[]};
+
+const store = new AsyncLocalStorage<AsyncState>();
 
 class ContextManager {
-  get current(): Context {
-    return store.getStore() || new SpanContext();
+  get asyncState(): AsyncState {
+    let asyncState = store.getStore();
+
+    if (asyncState === undefined) {
+      asyncState = {context: new SpanContext(), spans: []};
+      store.enterWith(asyncState);
+    }
+
+    return asyncState;
   }
 
-  withContext(callback: (...args: any[]) => void, ...args: any[]) {
-    return store.run(new SpanContext(), callback);
+  get current(): Context { return this.asyncState.context; }
+  get spans(): Span[] { return this.asyncState.spans; }
+
+  spansDup(): Span[] {
+    let asyncState = store.getStore();
+
+    if (asyncState === undefined) {
+      asyncState = {context: new SpanContext(), spans: []};
+    } else {
+      asyncState = {context: asyncState.context, spans: [...asyncState.spans]};
+    }
+
+    store.enterWith(asyncState);
+
+    return asyncState.spans;
+  }
+
+  clear(): void {
+    store.enterWith(undefined as unknown as AsyncState);
+  }
+
+  restore(context: Context, spans: Span[]): void {
+    store.enterWith({context, spans: spans || []});
+  }
+
+  withSpan(span: Span, callback: (...args: any[]) => any, ...args: any[]): any {
+    if(!span.startTime)
+      span.start();
+
+    try {
+      return callback(span, ...args);
+    } catch (e) {
+      span.error(e);
+      throw e;
+    } finally {
+      span.stop();
+    }
+  }
+
+  withSpanNoStop(span: Span, callback: (...args: any[]) => any, ...args: any[]): any {
+    if(!span.startTime)
+      span.start();
+
+    try {
+      return callback(span, ...args);
+    } catch (e) {
+      span.error(e);
+      span.stop();
+      throw e;
+    }
   }
 }
 
--- a/src/agent/Buffer.ts
+++ b/src/agent/Buffer.ts
@@ -20,6 +20,7 @@
 import { createLogger } from '../logging';
 import Segment from '../trace/context/Segment';
 import config from '../config/AgentConfig';
+import TraceReportClient from '../agent/protocol/grpc/clients/TraceReportClient';
 
 const logger = createLogger(__filename);
 
@@ -42,6 +43,7 @@ class Buffer {
       return this;
     }
     this.buffer.push(segment);
+    TraceReportClient.ref();  // this is currently hard-coded for grpc, if other protocols added need to change
 
     return this;
   }
--- a/src/agent/protocol/grpc/clients/TraceReportClient.ts
+++ b/src/agent/protocol/grpc/clients/TraceReportClient.ts
@@ -31,6 +31,7 @@ const logger = createLogger(__filename);
 
 class TraceReportClient implements Client {
   reporterClient: TraceSegmentReportServiceClient;
+  timeout: any;
 
   constructor() {
     this.reporterClient = new TraceSegmentReportServiceClient(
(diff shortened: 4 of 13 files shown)


tom-pytel:
Better, but still far from done, added:
* async() will temporarily remove span from parent spans list in parent async context.
* resync() will re-add to spans list in new child async context.

tom-pytel:
Configuration error I think, I don't get the "no-shadowed-variable  Shadowed name: ?" error locally when I build or lint.

kezhenxu94:
> Configuration error I think, I don't get the "no-shadowed-variable Shadowed name: ?" error locally when I build or lint.

What IDE are you using? I'm using WebStorm, make sure to enable tslint and prettier, or run `npm run lint` in the command line, also please rebase on the master (I enable the listing on master but the PR seems to be stale

tom-pytel:
> What IDE are you using? I'm using WebStorm, make sure to enable tslint and prettier, or run `npm run lint` in the command line, also please rebase on the master (I enable the listing on master but the PR seems to be stale

No IDE, just good old-fashioned command-line `npm run lint` and `npm run build`.

kezhenxu94:
@tom-pytel is it ready to review/merge now?

tom-pytel:
No not quite ready to merge, need to show cases where this works and previous one doesn't and want to export `ContextManager` for user specified spans. Hoping for later today or tomorrow latest.

tom-pytel:
But if you want to have a quick look, questions / comments are always welcome.

kezhenxu94:
> No not quite ready to merge, need to show cases where this works and previous one doesn't and want to export `ContextManager` for user specified spans. Hoping for later today or tomorrow latest.

Take your time, feel free to ping me when it's ready, thanks in advance 🙇

src/trace/context/ContextManager.ts
```diff
@@ -18,19 +18,74 @@
...
+    } finally {
+      if (stopOnSuccess) {
+        span.stop();
+      }
+    }
+  }
+
+  // withContext(callback: (...args: any[]) => void, ...args: any[]) {
+  //   return store.run({context: new SpanContext(), spans: []}, callback);
+  // }
```
kezhenxu94, line 88: Maybe we can get rid of this `withSpan` now to eliminate the "callbacks", user codes become

```typescript

        const span = ContextManager.current.newExitSpan(operation, host).start();

        span.component = Component.HTTP;
        span.layer = SpanLayer.HTTP;
        span.tag(Tag.httpURL(host + pathname));

        const request = original.apply(this, arguments);

        span.extract().items.forEach((item) => {
          request.setHeader(item.key, item.value);
        });
```

instead of 

```typescript

        const [span, request]: [Span, ClientRequest] = ContextManager.withSpan(false,
            ContextManager.current.newExitSpan(operation, host).start(), (_span: Span, args) => {

          _span.component = Component.HTTP;
          _span.layer = SpanLayer.HTTP;
          _span.tag(Tag.httpURL(host + pathname));

          const _request = original.apply(this, args);

          _span.extract().items.forEach((item) => {
            _request.setHeader(item.key, item.value);
          });

          return [_span, _request];

        }, arguments);
```

I'm sorry it may be `withContext` that misled you to follow the same style 😂

src/trace/context/ContextManager.ts
```diff
@@ -18,19 +18,74 @@
...
+      store.enterWith(asyncState);
+    }
+
+    return asyncState;
+  }
+
+  get current(): Context { return this.asyncState.context; }
+  get spans(): Span[] { return this.asyncState.spans; }
+
+  spans_dup(): Span[] {
```
kezhenxu94, line 44: let's unify the naming style (camel case), if we want to use snake case later, we can have another discussion/refactor then

kezhenxu94 (commented):
Had a quick look and left some comments, will take a detailed look when it's ready

kezhenxu94:
> * Context.capture() and restore() will be made redundant / unnecessary by this?

Yes I believe they're unnecessary after this PR, feel free to remove them altogether in this PR or in a following PR if you like.

> * ContextManagere.withSpan() export to user?

Like what I commented, `ContextManagere.withSpan()` seems to be not that useful but just for catching exceptions? I don't want to add another callback level in the users codes.

> * Node 10 does not have AsyncLocalStorage, will need to provide custom substitute class for this (using async_hooks.createHook init and destroy to track parent-child relationships).

Let's tackle Node10 a little bit later after this PR. "using async_hooks.createHook init and destroy to track parent-child relationships" this is my original thought before coming across `AsyncLocalStorage`, but be careful, we need to evaluate the memory usage and performance impact when tracking the parent-child relationship, (does `AsyncLocalStorage` do so behind the scene?)

> * I had this question with the Python agent whether entry spans should always start with a new context but decided against it because maybe someone wants a previous local context for some debug or other reason to chain before the entry context? So I did the same here and disabled fresh context on Entry span but rather just let it use the normal inheritance chain, which should eventually give a new context unless the user specifically prevents it.

👍 

> * Has problem as Python agent with background updater not firing if application exits very quickly, should not exit until all data sent, probably should hook 'beforeExit' event and flush segments, should do same in Python.

I'd consider this as a bug (not sending segment when it exists very quickly).

Let's create some issues to track those aforementioned that are not to be addressed in this PR 😄

src/trace/context/ContextManager.ts
tom-pytel: Actually that is what I did at first but realized this wrapper could be very handy especially for users to avoid errors. Yes this is a callback but it is a cheap synchronous callback (just a function call), not the more expensive needs-its-own-async-context type of callback. Have a look at how I am using it in the examples I send up and if you do not think it is a net positive then I will remove.

src/trace/context/ContextManager.ts
tom-pytel: I try to follow the convention of whatever project I am working on but sometimes my brain just doesn't work properly, of course that should be camel-case.

tom-pytel:
> > * Context.capture() and restore() will be made redundant / unnecessary by this?
> 
> Yes I believe they're unnecessary after this PR, feel free to remove them altogether in this PR or in a following PR if you like.

Next or following PR, don't want to get too ahead.

> Like what I commented, `ContextManagere.withSpan()` seems to be not that useful but just for catching exceptions? I don't want to add another callback level in the users codes.

See examples I will send.

> Let's tackle Node10 a little bit later after this PR. "using async_hooks.createHook init and destroy to track parent-child relationships" this is my original thought before coming across `AsyncLocalStorage`, but be careful, we need to evaluate the memory usage and performance impact when tracking the parent-child relationship, (does `AsyncLocalStorage` do so behind the scene?)

Doesn't really matter what the performance hit is since otherwise SW will not be usable at all on Node 10 without this, so having the option of running slower vs. not running at all is an obvious choice.

> I'd consider this as a bug (not sending segment when it exists very quickly).
> 
> Let's create some issues to track those aforementioned that are not to be addressed in this PR

Feel free to create the issues so they don't get forgotten about.

tom-pytel:
Ready for merge I think, tests need to be changed since I added tags which weren't previously being returned. I couldn't really come up with a ready-to-use example vs. the previous version since it is so limited that it really only did one thing, but I did hack it to try with the sever and client examples I will post here and it work well at all, so yes this PR is an improvement.

Server example:
```julia
const {default: Agent, ContextManager} = require('skywalking');

Agent.start({
  serviceName: '',
  serviceInstance: '',
  collectorAddress: '',
  authorization: '',
  maxBufferSize: 1000,
});

const http = require('http');

let count = 0;

const server = http.createServer(function (req, res) {
  if (count++ & 1)
    res.writeHead(404, {'Content-Type': 'text/html'});

  res.write('Hello World...\n');

  res.end();
}).listen(8000);
```

Client example:
```julia
const {default: Agent, ContextManager} = require('skywalking');

Agent.start({
  serviceName: '',
  serviceInstance: '',
  collectorAddress: '',
  authorization: '',
  maxBufferSize: 1000,
});

var http = require('http');

callback = function(response) {
  var str = '';

  response.on('data', function (chunk) {
    str += chunk;
  });

  response.on('end', function () {
    console.log(str);
  });
}

ContextManager.withSpan(ContextManager.current.newLocalSpan('local'), () => {
  http.request({host: 'localhost', path: '/1', port: '8000', methof: 'GET'}, callback).end();
  http.request({host: 'localhost', path: '/2', port: '8000', methof: 'GET'}, callback).end();
  http.request({host: 'localhost', path: '/3', port: '8000', methof: 'GET'}, callback).end();
});

setTimeout(() => null, 2500);  // allow time for BG update
```

tom-pytel:
Not quite sure what is failing here...

tom-pytel:
I am seeing a `Request failed with status code 500`, is this an actual failure and not a test config thing? And if so how can I get the actual failing code to check?

kezhenxu94:
> I am seeing a `Request failed with status code 500`, is this an actual failure and not a test config thing? And if so how can I get the actual failing code to check?

It's an actual failure because the actual trace data is different from the expected (in the `expected.data.yaml` file)

There is the actual trace data
https://github.com/apache/skywalking-nodejs/pull/2/checks?check_run_id=1470200800#step:5:11660 （remove the quotes, `+`, `\n`, etc.) you can compare it with the `expected.data.yaml` to see what is different

tom-pytel:
The exit spans are supposed to receive the http.status.msg now but I guess I will look into that in the next PR.

tom-pytel:
Ok I am an idiot, I forgot that I added the host to the URL tag in entry spans and the exit spans are storing the status msg as they should, though entry spans are not for some reason even though the code looks like they should be.

src/trace/context/Context.ts
```diff
@@ -36,13 +35,13 @@ export default interface Context {
 
   stop(span: Span): boolean;
 
+  async(span: Span): void;
+
+  resync(span: Span): void;
+
```
wu-sheng, line 41: Could you consider to fill the comments of all methods on this interface? I think those would be helpful for newer contributors, as we plan to release 0.1.0.

kezhenxu94 (approved):
It is good enough for what this PR aims to solve, let's merge it for now, further polishing/documenting PRs can follow after this, before 0.1.0,

@tom-pytel thanks very much for the patience

PR #3, opened 2020-11-30
Clean up unused codes and polish

kezhenxu94:


.asf.yaml
```diff
@@ -30,7 +30,4 @@ github:
     squash:  true
     merge:   false
     rebase:  false
-  protected_branches:
-    master:
-      required_pull_request_reviews:
-        dismiss_stale_reviews: true
```
wu-sheng, line 36: Why removing this?

wu-sheng (commented):
So, you are beginning to hold the tracing context until the async finished? And remove the capture/restore(continued?)?
If so, how do you resolve the batch scheduled task? Which could cause long latency, and you have to hold the context for a very long time

kezhenxu94:
> So, you are beginning to hold the tracing context until the async finished? And remove the capture/restore(continued?)?
> 
> If so, how do you resolve the batch scheduled task? Which could cause long latency, and you have to hold the context for a very long time

What's the difference, IIUC, `capture`d context is also hold in memory until it's `restore`d/`continued`, did I understand your question correctly?

wu-sheng:
> What's the difference, IIUC, captured context is also hold in memory until it's restored/continued, did I understand your question correctly?

At least in the Java implementation, capture doesn't hold the whole context, mainly about finished spans, which could reduce unnecessary memory cost.

kezhenxu94:
> > What's the difference, IIUC, captured context is also hold in memory until it's restored/continued, did I understand your question correctly?
> 
> At least in the Java implementation, capture doesn't hold the whole context, mainly about finished spans, which could reduce unnecessary memory cost.

Compared with the Java's, the contexts in NodeJS are rather lightweight (due to the natively supported `AsyncLocalStorage`), which nearly only contains the finished spans, see this


https://github.com/apache/skywalking-nodejs/blob/2fa4cca1aed3b34c5fe17c15ed72a54403b8a853/src/trace/context/SpanContext.ts#L34-L37

wu-sheng:
Isn't `segment: Segment = new Segment();` including the finished spans?

PR #5, opened 2020-12-02
[Enhancement] tweaks for better error handling

tom-pytel:
* Removed ContextManager.withSpanNoStop() to clean up interface since this was only meant to be used in plugins and so the plugins themselves should do this.
* Changed ContextManager.withSpan() and .withSpanAsync() to not insert span as first argument but rather just pass the args through to allow more generalized use, for example `ContextManager.withSpanAsync(user_defined_span, unsupported_async_URL_requester, requester_arg1, ...)`.
* Reworked HttpPlugin:http.request() to better cover and handle error conditions, should be 100% covered now (before there was a range between `span.async()` and the callback where an unexpected error could cause the span to not be stopped).

### DIFF
--- a/src/plugins/HttpPlugin.ts
+++ b/src/plugins/HttpPlugin.ts
@@ -60,44 +60,49 @@ class HttpPlugin implements SwPlugin {
               };
         const operation = pathname.replace(/\?.*$/g, '');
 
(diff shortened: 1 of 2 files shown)


kezhenxu94 (approved):
Looks good

PR #6, opened 2020-12-03
tslib as normal not dev dependancy

tom-pytel:
This dependency missing if installed normally as npm package, so moved from dev to normal dependancies. If this is not the right way to do it then feel free to correct it some other way.


kezhenxu94 (approved):
Seems to be the case, thanks

PR #7, opened 2020-12-03
Node 10 AsyncLocalStorage and Plugin loader

tom-pytel:
- For the project I am working on Node 10 support is a requirement so I had to do this class to replicate AsyncLocalStorage functionality. It is only used if AsyncLocalStorage does not already exist in async_hooks, otherwise SW would error out under Node 10.
- Fixed plugin loader to actually load plugins. :smiley: *(which are not builtin)*

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,7 +29,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [ 12, 14 ]
+        node-version: [ 10, 12, 14 ]
     steps:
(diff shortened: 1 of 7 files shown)


kezhenxu94:
I added Node 10 to the test matrix to verify this works


> The commits will be squashed when merging and you are the only author of this pr

kezhenxu94:
I finally realized that the tests are running inside the docker, so the matrix doesn't make much sense, I'll merge this first and open another PR to set up tests under different Node versions

kezhenxu94 (approved):

PR #9, opened 2020-12-04
[WIP] express plugin, make spans actually work

tom-pytel:
* Fixed nested Entry and Exit spans, before they did not at all.
* Fixed DummySpan - override `inject()` and `extract()`.
* Fixed PluginInstaller - previously did not load non-builtin modules correctly, also now wraps plugin install in try / catch.
* Fixed HttpPlugin - now storing peer.
* Changed `StackedSpan` to track nested depth more clearly through `start()` and `stop()` instead of `finish()`.
* Changed `inject()` and `extract()` functions to have the same meaning as Python agent.
* Added https plugin.
* Added express plugin.
* Added axios plugin.

### DIFF
--- a/src/trace/context/SpanContext.ts
+++ b/src/trace/context/SpanContext.ts
@@ -17,8 +17,11 @@
  *
  */
 
+import config from '../../config/AgentConfig';
 import Context from '../../trace/context/Context';
+import DummyContext from '../../trace/context/DummyContext';
 import Span from '../../trace/span/Span';
+import DummySpan from '../../trace/span/DummySpan';
 import Segment from '../../trace/context/Segment';
 import EntrySpan from '../../trace/span/EntrySpan';
 import ExitSpan from '../../trace/span/ExitSpan';
@@ -50,6 +53,18 @@ export default class SpanContext implements Context {
     return this.parent ? this.parent.id : -1;
   }
 
+  ignoreCheck(operation: string, type: SpanType): Span | undefined {
+    if (operation.match(config.reIgnoreOperation)) {
+      return new DummySpan({
+        context: new DummyContext(),
+        operation: '',
+        type,
+      });
+    }
+
+    return undefined;
+  }
+
   newEntrySpan(operation: string, carrier?: ContextCarrier): Span {
     if (logger.isDebugEnabled()) {
       logger.debug('Creating entry span', {
@@ -58,10 +73,13 @@ export default class SpanContext implements Context {
       });
     }
 
+    let span = this.ignoreCheck(operation, SpanType.ENTRY);
+
+    if (span)
+      return span;
+
     const spans = ContextManager.spansDup();
     const parent = spans[spans.length - 1];
-
-    let span;
 
     if (parent && parent.type === SpanType.ENTRY) {
       span = parent;
@@ -91,10 +109,13 @@ export default class SpanContext implements Context {
       });
     }
 
+    let span = this.ignoreCheck(operation, SpanType.EXIT);
+
+    if (span)
+      return span;
+
     const spans = ContextManager.spansDup();
     const parent = spans[spans.length - 1];
-
-    let span;
 
     if (parent && parent.type === SpanType.EXIT) {
       span = parent;
@@ -123,6 +144,11 @@ export default class SpanContext implements Context {
         executionAsyncId: executionAsyncId(),
       });
     }
+
+    let span = this.ignoreCheck(operation, SpanType.LOCAL);
+
+    if (span)
+      return span;
 
     ContextManager.spansDup();
 
--- a/src/trace/context/SpanContext.ts
+++ b/src/trace/context/SpanContext.ts
@@ -30,6 +30,7 @@ import Snapshot from '../../trace/context/Snapshot';
 import SegmentRef from '../../trace/context/SegmentRef';
 import { ContextCarrier } from './ContextCarrier';
 import ContextManager from './ContextManager';
+import { SpanType } from '../../proto/language-agent/Tracing_pb';
 
 const logger = createLogger(__filename);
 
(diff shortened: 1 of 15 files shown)


tom-pytel:
Config or resource issue?

kezhenxu94:
> Config or resource issue?

Network from GitHub Actions runtime to hub.docker is unstable, have rerun it

kezhenxu94:
> Also, why are the meanings of Span.inject() and Span.extract() reversed in this agent compared to the Python one?

This is a stupid mistake I made, it should have had the same meanings as the Python agent’s. 

> Would be good to have a chat, this agent was in very poor shape and I could use some clarification to make it work. 

This agent indeed lacked of a lot of efforts, would be happy to have contributors like you to help to make it better. Feel free to ask questions in Slack channel or GitHub issues here, or mailing list. Thanks

tom-pytel:
> This agent indeed lacked of a lot of efforts, would be happy to have contributors like you to help to make it better. Feel free to ask questions in Slack channel or GitHub issues here, or mailing list. Thanks

Which one will give me the quickest replies?

kezhenxu94:
> > This agent indeed lacked of a lot of efforts, would be happy to have contributors like you to help to make it better. Feel free to ask questions in Slack channel or GitHub issues here, or mailing list. Thanks
> 
> 
> 
> Which one will give me the quickest replies?

I check GitHub notifications most frequently, and reply as soon as I saw them, but due to the different time zones, I might wake up to check them in hours after your posts.

tom-pytel:
> I check GitHub notifications most frequently, and reply as soon as I saw them, but due to the different time zones, I might wake up to check them in hours after your posts.

Ok, I will just keep posting here then.

tom-pytel:
`2020-12-05T12:43:00.079Z testcontainers ERROR Failed to start DockerCompose environment: Creating network "testcontainers-28a887eed5d9791e474a4847a6654043_traveling-light" with the default driver`

tom-pytel:
Well the code for this done, it still needs a test and a Component ID allocated.

kezhenxu94:
> Well the code for this done, it still needs a test and a Component ID allocated.

The Component ID is allocated [here](https://github.com/apache/skywalking/blob/83dd23977dff90f22a7bea30b04bcfcfc013e8ec/oap-server/server-bootstrap/src/main/resources/component-libraries.yml#L455-L478)

tom-pytel:
Hmm, express was already there so I will just use that I guess. As for tests, I think it is something that would maybe take you 5 minutes but could take me hours to figure out your setup, any chance you could do the test for this?

src/trace/context/SpanContext.ts
```diff
@@ -57,39 +58,62 @@ export default class SpanContext implements Context {
...
+      span = new ExitSpan({
+        id: this.spanId++,
+        parentId: this.parentId,
+        context: this,
+        peer,
+        operation,
+      });
+
+      // if (carrier && carrier.isValid()) {  // is this right?
+      //   Object.assign(carrier, span.inject());
```
tom-pytel, line 112: Forgot this detail, what is the proper way to do the injection in `SpanContext.newExitSpan()`?

tom-pytel:
I did not differentiate https from http in any way in terms of identifying itself as http and it lives in the same module because it is literally the same code just applied to a different built-in module.

kezhenxu94:
e

> I did not differentiate https from http in any way in terms of identifying itself as http and it lives in the same module because it is literally the same code just applied to a different built-in module.

+1 from me

> 
> Also, I noticed there exists a specific component definition for NodeJS http server:
> 
> ```yaml
> HttpServer:
>   id: 4001
>   languages: Node.js
> ```

This was originally for http://github.com/SkyAPM/SkyAPM-nodejs/ 

> But the one actually used in the http plugin is #49:
> 
> ```yaml
> http:
>   id: 49
>   languages: Java,C#,Node.js
> ```
> 
> And the client used is # 2.
> Which way do you guys want to go with this, more specific for the individual languages or should all language-common base protocols be the same (# 2 and # 49)?

Basically the latter one applies. The component IDs are only used to match the logo/icon in the web UI, if there is no logo/icon for that (e.g. most built-in modules), we just reuse the existing ID from Java language that shares the same protocol (e.g, http)

src/trace/context/SpanContext.ts
kezhenxu94: You don't need to do anything here, context carriers from exit span are meant to be sent to remote peer, so the plugin (or user codes) should tackle this and propagate them to remote peer (via http headers or MQ metadata, etc.) properly, like the http plugin

```nodejs
span.inject().items.forEach((item) => {
   request.setHeader(item.key, item.value);
});
```

kezhenxu94:
> As for tests, I think it is something that would maybe take you 5 minutes but could take me hours to figure out your setup, any chance you could do the test for this?

Yes I can help on this, when you think it's all set except for the test, ping me I'll push to this branch directly to verify the plugin. Thanks :)

src/trace/context/SpanContext.ts
tom-pytel: I was going off the Python agent, it does something like this:
```python
    def new_exit_span(self, op: str, peer: str, carrier: 'Carrier' = None) -> Span:
        span = self.ignore_check(op, Kind.Exit)
        if span is not None:
            return span

        spans = _spans_dup()
        parent = spans[-1] if spans else None  # type: Span

        span = parent if parent is not None and parent.kind.is_exit else ExitSpan(
            context=self,
            sid=self._sid.next(),
            pid=parent.sid if parent else -1,
            op=op,
            peer=peer,
        )

        if carrier is not None:
            span.inject(carrier=carrier)

        return span
```

tom-pytel:
> Yes I can help on this, when you think it's all set except for the test, ping me I'll push to this branch directly to verify the plugin (or you'd like to merge this first, just let me know). Thanks :)

Express and https should be good to go so do in whatever order you thinks makes sense.

src/trace/context/SpanContext.ts
kezhenxu94: Hmm, sorry I was wrong, let me check a little more

tom-pytel:
Some small fixes here but mostly added suffix and path ignore mechanism via `SW_IGNORE_SUFFIX` and `SW_TRACE_IGNORE_PATH`. There is no `SW_TRACE_IGNORE` like in the Python agent, instead the presence of a non-empty `SW_TRACE_IGNORE_PATH` enables path ignore. I will probably change it in the Python agent to work this was if you have no objections (cleaner, fewer env vars to deal with).

tom-pytel:
This interacts with http plugin below it just like everything else, and event though they behave well together I am not completely happy with how this works right now. I think it can almost certainly be made more efficient and the interactions better defined (needs testing).

tom-pytel:
> > Yes I can help on this, when you think it's all set except for the test, ping me I'll push to this branch directly to verify the plugin (or you'd like to merge this first, just let me know). Thanks :)
> 
> Express and https should be good to go so do in whatever order you thinks makes sense.

Ok, well I am getting a little farther ahead than I thought and this PR is getting bigger, as well as a parallel branch I have for our own stuff which is getting more and more different. So maybe it would be better if you could merge this PR before anything else?

kezhenxu94 (approved):

kezhenxu94:
> > > Yes I can help on this, when you think it's all set except for the test, ping me I'll push to this branch directly to verify the plugin (or you'd like to merge this first, just let me know). Thanks :)
> > 
> > 
> > Express and https should be good to go so do in whatever order you thinks makes sense.
> 
> Ok, well I am getting a little farther ahead than I thought and this PR is getting bigger, as well as a parallel branch I have for our own stuff which is getting more and more different. So maybe it would be better if you could merge this PR before anything else?

Sure. Thanks.

PR #12, opened 2020-12-15
Init the client twoice,Make sure the config take effect

QuanjieDeng:
This may not be the optimal solution, but at present I can only modify it like this

### DIFF
--- a/src/agent/protocol/grpc/clients/HeartbeatClient.ts
+++ b/src/agent/protocol/grpc/clients/HeartbeatClient.ts
@@ -33,23 +33,14 @@ import * as os from 'os';
 const logger = createLogger(__filename);
 
 class HeartbeatClient implements Client {
-  heartbeatClient: ManagementServiceClient;
+  heartbeatClient?: ManagementServiceClient;
   heartbeatTimer?: NodeJS.Timeout;
 
-  constructor() {
-    this.heartbeatClient = new ManagementServiceClient(config.collectorAddress, grpc.credentials.createInsecure(), {
-      interceptors: [AuthInterceptor],
-    });
-  }
-
   get isConnected(): boolean {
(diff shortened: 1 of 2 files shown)


src/agent/protocol/grpc/clients/HeartbeatClient.ts
```diff
@@ -47,6 +47,13 @@ class HeartbeatClient implements Client {
   }
 
   start() {
+    /*
+    init  the heartbeatClient again
+    */
```
kezhenxu94, line 52: Remove this comment and the initialization in the `constructor`

src/agent/protocol/grpc/clients/TraceReportClient.ts
```diff
@@ -50,6 +50,14 @@ class TraceReportClient implements Client {
   }
 
   start() {
+    /*
+    init the reporterClient again
+    */
```
kezhenxu94, line 55: Remove this comment and the initialization in the `constructor`

kezhenxu94:
Another workaround is to set the config via environment variables

kezhenxu94:
Hey @QuanjieDeng , why do you close the PR, the fix is correct, I just wanted you to remove the duplicated initialization in the constructors

QuanjieDeng:
> Hey @QuanjieDeng , why do you close the PR, the fix is correct, I just wanted you to remove the duplicated initialization in the constructors

yes,Im not good at TypeScript,Im trying  fix this  bug

kezhenxu94:
> > Hey @QuanjieDeng , why do you close the PR, the fix is correct, I just wanted you to remove the duplicated initialization in the constructors
> 
> yes,Im not good at TypeScript,Im trying fix this bug

It's OK

kezhenxu94 (approved):

PR #14, opened 2020-12-20
Update submodule over https instead of ssh in case of no ssh is configured

kezhenxu94:
Developers/CI machines where no ssh is configured, it failed to update the submodule


wu-sheng (approved):
https is better.

PR #15, opened 2020-12-25
Make the contract channels clear

wu-sheng:
Do we need to process release now?


kezhenxu94:
> Do we need to process release now?

I'll start the release process one by one after the Python's, not to mess up the mailing list

kezhenxu94 (approved):

PR #22, opened 2021-01-12
add context clear before new entry span

zijin-m:
Different requests are assigned to the same traceId causing data errors
image

Because request use old asyncState
image

So,I Add `clear` before `newEntrySpan`, Is there a better solution, or am I misunderstanding？

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -516,6 +516,31 @@
         "minimist": "^1.2.0"
       }
     },
+    "@grpc/grpc-js": {
+      "version": "1.2.3",
+      "resolved": "http://npm.meshkit.cn/@grpc%2fgrpc-js/-/grpc-js-1.2.3.tgz",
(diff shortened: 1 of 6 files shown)


wu-sheng:
> So,I Add clear before newEntrySpan, Is there a better solution, or am I misunderstanding？

If the context is not cleared, then this should be a bug. The entry span doesn't mean the first span.

zijin-m:
> > So,I Add clear before newEntrySpan, Is there a better solution, or am I misunderstanding？
> 
> If the context is not cleared, then this should be a bug. The entry span doesn't mean the first span.


Yes, after my test, it seems that `AsyncLocalStorage` did not recreate the async context for the new request event, which caused `getStore` to obtain the asyncState of the **previous request event**, which may be related to the browser’s TCP Connection Reuse, because when I use ab or curl To test and there is no such problem

wu-sheng:
@kezhenxu94 may know more about the tech detail, but basically this fix seems not correct to me.

PR #28, opened 2021-01-27
Added http protocol to plugins

tom-pytel:


tom-pytel:
Ok, I really do not understand, I changed the fields affected by the PR in the expected.data.yaml files and it still fails. Comparing the "actualData" from the failure against the expected data it all matches up except for some "isError: false" which are not present in expected.data.yaml, but that yaml without the "isError" is coming from the previous successful master which I assume is ok. So for future pushes so that I can get the test right can you please show me what I am missing in order to get the test to succeed?

kezhenxu94:
> Comparing the "actualData" from the failure against the expected data it all matches up except for some "isError: false" which are not present in expected.data.yaml, but that yaml without the "isError" is coming from the previous successful master which I assume is ok.

That is what we're doing to find out what's wrong.

> So for future pushes so that I can get the test right can you please show me what I am missing in order to get the test to succeed?

May be more carefully when comparing? 🤣 I compare those and see the orders of the tags are different, so please always keep in mind the verifier is not that flexible to ignore the orders

tom-pytel:
> > Comparing the "actualData" from the failure against the expected data it all matches up except for some "isError: false" which are not present in expected.data.yaml, but that yaml without the "isError" is coming from the previous successful master which I assume is ok.
> 
> That is what we're doing to find out what's wrong.
> 
> > So for future pushes so that I can get the test right can you please show me what I am missing in order to get the test to succeed?
> 
> May be more carefully when comparing?  I compare those and see the orders of the tags are different, so please always keep in mind the verifier is not that flexible to ignore the orders

:facepalm:

tom-pytel:
> May be more carefully when comparing?  I compare those and see the orders of the tags are different, so please always keep in mind the verifier is not that flexible to ignore the orders

Ok wait a minute, but I didn't change the order of tags, those were the .yaml files from a previous SUCCESSFUL master and all I did was change the url VALUE, didn't add or remove any other fields.

src/plugins/ExpressPlugin.ts
```diff
@@ -95,6 +94,8 @@ class ExpressPlugin implements SwPlugin {
       } catch (e) {
         stopIfNotStopped(e);
         throw e;
+      } finally {  // req.protocol is only possibly available after call to _handle()
+        span.tag(Tag.httpURL(((req as any).protocol ? (req as any).protocol + '://' : '') + (req.headers.host || '') + req.url));
```
kezhenxu94, line 98: You moved this, is this the reason?

tom-pytel: Yes, because that field `req.protocol` is not available until after the call to `_handle()`.

kezhenxu94: > Yes, because that field `req.protocol` is not available until after the call to `_handle()`.

I meant this is the reason why the tag orders changed. Is it?

tom-pytel: :facepalm: :facepalm: :facepalm:

kezhenxu94 (approved):
Thanks

PR #29, opened 2021-01-28
add plugin mysql

zijin-m:
In addition, use the `resolve` package to read `package.json` because `mysql2` does not add `package.json` in the `exports` field of the `package.json` file


src/core/PluginInstaller.ts
```diff
@@ -33,7 +34,7 @@ while (topModule.parent) {
 export default class PluginInstaller {
   private readonly pluginDir: string;
   readonly require: (name: string) => any = topModule.require.bind(topModule);
-  private readonly resolve = (request: string) => (module.constructor as any)._resolveFilename(request, topModule);
+  private readonly resolve = (request: string) => resolve.sync(request);
```
tom-pytel, line 37: Question about this, the current form `(module.constructor as any)._resolveFilename(request, topModule);` ensures that the resolution is done from the point of view of the topmost module (the main app). This is necessary because if for example you have skwalking `npm link`ed into a project instead of installed alongside all its dependancies then normal resolve will get the wrong path, one that sits in the skywalking node_modules path instead of the applicatiuon. Does the `resolve` package do this?

tom-pytel:
I can look into adding tests later, right now fighting with a PostgreSQL plugin. If you can merge this before I get the `pg` plugin done it will make things less messy, especially with the other `mysql2` plugin...

tom-pytel:
> In addition, use the `resolve` package to read `package.json` because `mysql2` does not export the file `package.json`

What do you mean by "does not export"? It HAS a `package.json`.

src/core/PluginInstaller.ts
zijin-m: Sorry, I missed this point. According to the documentation [resolve](https://www.npmjs.com/package/resolve), this can be done by 
```private readonly resolve = (request: string) => resolve.sync(request, {basedir: path.dirname(topModule.id) });```
 if You agree i will update later

zijin-m:
> > In addition, use the `resolve` package to read `package.json` because `mysql2` does not export the file `package.json`
> 
> What do you mean by "does not export"? It HAS a `package.json`.

According to the node documentation [modules](https://nodejs.org/api/esm.html#esm_subpath_exports) 

``` Bare specifiers like 'some-package' or 'some-package/shuffle'. They can refer to the main entry point of a package by the package name, or a specific feature module within a package prefixed by the package name as per the examples respectively. Including the file extension is only necessary for packages without an "exports" field.```

 while `mysql2` defines the `exports` field in its  [package.json](https://github.com/sidorares/node-mysql2/blob/master/package.json#L45) file, similar to [issue](https://github.com/react-native-community/cli/issues/1168#issuecomment-628034263)

src/core/PluginInstaller.ts
tom-pytel: Ok, well I checked your fix and it seems to work, so I am ok with this.

tom-pytel:
Some ideas for your plugin:
* Wrap entire section after `newExitSpan()` with try / catch so that you can catch any exception from `_query.apply()` or even your own code and `span.error()` that and then `span.stop()` cleanly.
* Put  static init stuff like `span.component = Component.MYSQL` before the query call so that if there is an exception the span will have as much information as possible.
* Does `mysql2` do row streaming like `mysql`? If so it should be handled.
* And a really silly detail, you don't need `span.errored = true` before `span.error(err)`, that is set by the `.error()` function.

@kezhenxu94 asked me to review this because I also recently added a `mysql` plugin, just the first one, not `mysql2`, so you can to look at that for some ideas of what I am talking about, its in master.

tom-pytel:
Actually, better look at https://github.com/apache/skywalking-nodejs/pull/31, master had some errors.

tom-pytel:
Hey @zijin-m, going back to #27 where you said:

> After testing, if two requests are repeatedly sent in a short time, the subsequent requests may use the AsyncState initialized by the previous request. It is temporarily unclear why store.enterWith(undefined as unknown as AsyncState) in the clear method not work(The store cannot be reset), I tried to add an invalid flag, which works normally, I hope to discuss the feasibility of this implementation with you, thx

I realize some time has passed, but is there any chance you can provide a short snippet case of this happening before the `invalid` flag fix? I am looking at some other things regarding async plugins living together peacefully.

kezhenxu94:
Closing in favor of #54

PR #30, opened 2021-02-22
Add MySQLPlugin to plugins

tom-pytel:


src/config/AgentConfig.ts
```diff
@@ -59,5 +60,6 @@ export default {
     Number.parseInt(process.env.SW_AGENT_MAX_BUFFER_SIZE as string, 10) : 1000,
   ignoreSuffix: process.env.SW_IGNORE_SUFFIX ?? '.jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg',
   traceIgnorePath: process.env.SW_TRACE_IGNORE_PATH || '',
+  mysql_sql_parameters_max_length: Math.trunc(Math.max(0, Number(process.env.SW_MYSQL_SQL_PARAMETERS_MAX_LENGTH))) || 512,
```
kezhenxu94, line 63: Add the env var to doc please

kezhenxu94 (commented):
Add the plugin to the doc https://github.com/apache/skywalking-nodejs#supported-libraries please

tom-pytel:
BTW, I've checked packages `request`, `request-promise` and `koa` and as expected they work by default through the underlying `http` layer. Should such packages be listed in the docs as "supported" or "compatible" because of this?

kezhenxu94:
> Should such packages be listed in the docs as "supported" or "compatible" because of this?

I think the latter saying makes more sense

tom-pytel:
> > Should such packages be listed in the docs as "supported" or "compatible" because of this?
> 
> I think the latter saying makes more sense, let's add a column in the supported libraries table

Go ahead and add the column how you want and when I test packages I will add them there.

kezhenxu94:
@tom-pytel can you please also review https://github.com/apache/skywalking-nodejs/pull/29 , since you're adding 2 plugins for 2 different MySQL libs. Also, if possible, add a test case in this PR?

tom-pytel:
I can look into adding tests later, right now fighting with a PostgreSQL plugin. If you can merge this before I get the pg plugin done it will make things less messy, especially with the other mysql2 plugin...

kezhenxu94 (approved):

PR #36, opened 2021-03-04
Fixing async and preparing child entry/exit

tom-pytel:
The changes here are the result of preparing for allowing Exit/Exit and Entry/Entry parent child relationships during which I discovered and am in the process of fixing and improving some async functionality.

The source of the problem which necessitated #27 was discovered and fixed. The `http` and `axios` modules did not properly `async()` on returning a connection. Also, `async()` itself now duplicates the spans list for the current task in order that async tasks created before the `async()` don't have their copies of the list corrupted by subsequent operations in the current task. This all removes the necessity for a separate `invalid` flag, which flag itself was only a partial patch for the main problem which would still manifest in other ways.

The functionality of the `http` module is also being expanded so that errors on `'data'` events are properly attributed to the span.

Also, `span.resync()` no longer needed before `span.stop()`, a span can be errored or stopped while async.

Not for merge just yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ Environment Variable | Description | Default
 | `SW_AGENT_INSTANCE` | The name of the service instance | Randomly generated |
 | `SW_AGENT_COLLECTOR_BACKEND_SERVICES` | The backend OAP server address | `127.0.0.1:11800` |
 | `SW_AGENT_AUTHENTICATION` | The authentication token to verify that the agent is trusted by the backend OAP, as for how to configure the backend, refer to [the yaml](https://github.com/apache/skywalking/blob/4f0f39ffccdc9b41049903cc540b8904f7c9728e/oap-server/server-bootstrap/src/main/resources/application.yml#L155-L158). | not set |
-| `SW_AGENT_LOGGING_LEVEL` | The logging level, could be one of `CRITICAL`, `FATAL`, `ERROR`, `WARN`(`WARNING`), `INFO`, `DEBUG` | `INFO` |
+| `SW_AGENT_LOGGING_LEVEL` | The logging level, could be one of `error`, `warn`, `info`, `debug` | `info` |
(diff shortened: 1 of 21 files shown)


tom-pytel:
Ok, this PR is ready. To reiterate, the point of this plugin inheritance scheme is to make sure things like:
```js
app.get('/mysqlexit', function (req, res, next) {
  http.request(`http://somewhere.com/`, (r) => {
    mysqlConnection.query('SELECT * FROM `table`', function (error, results, fields) {
      if (error)
        next(error);
      else
        res.end(JSON.stringify(results));
    });
  }).end();
});
```
work correctly. Previously the `http.request()` and `mysqlConnection.query()` exit spans would be merged into a single one, now the db exit span will be a child of the http exit span since it is created in user code which is actually processing the http request.

With something like Axios which stops its span before passing control back to the user:
```js
app.get('/mysqlexit', function (req, res, next) {
  axios.get(`http://localhost:${portOut}/`).then((r) => {
    mysqlConnection.query('SELECT * FROM `table`', function (error, results, fields) {
      if (error)
        next(error);
      else
        res.end(JSON.stringify(results));
    });
  });
});
```
The db exit span will be created after the http span finishes and so will be a sibling of the http span and a child of the `app.get()` entry span.

tom-pytel:
Hold off on merging a bit, still working out some details.

tom-pytel:
I restored this because apart from covering up some of the errors in async before this flag was doing one other important thing that didn't have to do with those errors.

tom-pytel:
Ok, NOW it looks good, any other errors that might pop up can be fixed in other PRs.

tom-pytel:
> Tests under Node 10 failed, although I've rerun it, still failed, can you recheck?

Its a legitimate problem, not just config, something that's different between N10 and 12/14 is causing the spans to not stop in http, looking at it.

tom-pytel:
Ok, well this is in a good-ish place now, all my tests are running correctly. I will probably do more work on this but maybe this would be a good place to checkpoint and merge as this PR is getting a little big and already has several different feature changes?

kezhenxu94 (approved):

PR #38, opened 2021-03-12
Refactor and cleanups

tom-pytel:
Centralizing some functionality, minor fixes and cleanups, not for merge yet, will advise.

### DIFF
--- a/src/annotations/index.ts
+++ b/src/annotations/index.ts
@@ -28,7 +28,9 @@ export function trace(op?: string) {
     const original = descriptor.value;
 
     descriptor.value = function (...args: any[]) {
-      const span = ContextManager.current.newLocalSpan(op || key).start();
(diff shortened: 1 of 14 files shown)


tom-pytel:
* Changed Axios hook to higher level which will include all middleware in span time as well as allow adapters other than http.
* Changed how http client request works if it is used by a higher level plugin (like Axios) - now it just does nothing to avoid duplicating work that the higher level plugin does and allows higher level plugin to work if http plugin is disabled.

tom-pytel:
This PR makes that express can work if http plugin is disabled and vice-versa and also that express will not duplicate http plugin work if that is enabled.

tom-pytel:
I think this is good to merge here.

kezhenxu94 (approved):

PR #41, opened 2021-03-23
fix child spans created on immediate cb from op

tom-pytel:
Two MongoDB fixes here:
* Added flag to indicate whether mongodb is inside an internal call to be able to differentiate when a recursed operation belongs to mongo or the user so that user spans are properly created in case of an immediate callback from mongo but not in case of recursed mongo operations.
* mapReduce will show string functions quoted to differentiate from real passed functions.

### DIFF
--- a/src/plugins/MongoDBPlugin.ts
+++ b/src/plugins/MongoDBPlugin.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import SwPlugin, {wrapPromise} from '../core/SwPlugin';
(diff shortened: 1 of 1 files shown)


tom-pytel:
Something strange is going on with the tests, first the Node 10 test failed but the others succeeded, now Node 10 and 12 successful but Node 14 failed for a strange reason:
```
npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
```

tom-pytel:
Looks like it fixed itself.

kezhenxu94 (approved):

PR #44, opened 2021-03-26
Node Mongoose Plugin

tom-pytel:
Too bad you already released 0.2.0, I wanted to get this last plugin in before that. It is not done yet, still need to test some more  and write automated test. Also need to add the Component ID in the main skywalking repository and I think a logo goes somewhere?


wu-sheng:
> Too bad you already released 0.2.0

We could have 0.3.0 when need :) Releasing one agent is not relying on many things.

tom-pytel:
BTW, this is good for merge.

kezhenxu94 (commented):
The `package-lock.json` doesn't seem to be updated correctly, (seems it contains the dependencies of skywalking-nodejs itself and thus its depdnecneies duplicately), I simply checked out this PR and run `npm install`, the `package-lock.json` file is much more reduced, ~10000 lines!!), can you check ?

tom-pytel:
> The `package-lock.json` doesn't seem to be updated correctly, (seems it contains the dependencies of skywalking-nodejs itself and thus its depdnecneies duplicately), I simply checked out this PR and run `npm install`, the `package-lock.json` file is much more reduced, ~10000 lines!!), can you check ?

My npm probably mangled it when I added mongoose as a dev dependancy for the test.

tom-pytel:
If you want just use the previous good version of package-lock and add mongoose as a dependency yourself or if you want I try again and hopefully npm behaves.

tom-pytel:
Confirmed, my npm v7.6.3 keeps doing this.

kezhenxu94 (approved):

zhaozhiming:
Hi, I found a problem, when I add skywalking agent in my server, the mongoose api is broken. The `find` api is ok but it return a `Promise` but not a mongoose model. when I remove the agent everything is fine. 
I see that this case is not included in the test case, maybe it could be added to it.

```js
const result = await this.model.find({}).exec();
//  ERROR 57176 nodejs.TypeError: model.find(...).exec is not a function 
```

kezhenxu94:
> Hi, I found a problem, when I add skywalking agent in my server, the mongoose api is broken. The `find` api is ok but it return a `Promise` but not a mongoose model. when I remove the agent everything is fine.
> I see that this case is not included in the test case, maybe it could be added to it.
> 
> ```js
> const result = await this.model.find({}).exec();
> //  ERROR 57176 nodejs.TypeError: model.find(...).exec is not a function 
> ```


@tom-pytel mind taking a look?

tom-pytel:
> Hi, I found a problem, when I add skywalking agent in my server, the mongoose api is broken. The `find` api is ok but it return a `Promise` but not a mongoose model. when I remove the agent everything is fine.
> I see that this case is not included in the test case, maybe it could be added to it.
> 
> ```js
> const result = await this.model.find({}).exec();
> //  ERROR 57176 nodejs.TypeError: model.find(...).exec is not a function 
> ```

Ok well right off the bat I think I see the problem, the returned promise is a wrapped promise and `exec()` is not instrumented on that. Fixable, will do.

abdusamadtv:
@tom-pytel excuse, but do you have any news about this bug?

still not working and throwing this error

`MonitoringUser.find(...).populate is not a function`

image

tkNobug:

tom-pytel:
> @tom-pytel excuse, but do you have any news about this bug?

Yeah, I never got back to this. The reason for the failure is because the `find()` function returns a wrapped `Cursor` object which does not have those specific functions being handled. If you want to look at it yourself and see if you can add those you can look in `src/plugins/MongoDBPlugin.ts` method `hookCursorMaybe()`.

abdusamadtv:
Thanks for your response. Are you planning any plugin updates in the future?

abdusamadtv:
> > @tom-pytel excuse, but do you have any news about this bug?
> 
> Yeah, I never got back to this. The reason for the failure is because the `find()` function returns a wrapped `Cursor` object which does not have those specific functions being handled. If you want to look at it yourself and see if you can add those you can look in `src/plugins/MongoDBPlugin.ts` method `hookCursorMaybe()`.

Thanks for your response. Are you planning any plugin updates in the future?

tom-pytel:
> Thanks for your response. Are you planning any plugin updates in the future?

Not in the immediate future, sorry.

amolngt:
can anyone tell me which version should i use to avoid error ?
```
{
  message: 'Error installing plugin mongodb *',
  level: 'error',
  file: '/repo_name/node_modules/skywalking-backend-js/lib/core/PluginInstaller.js'
}
```

i am currently using "skywalking-backend-js": "^0.7.0" .

Thank you in advance.

PR #52, opened 2021-05-17
Cold start detection and marking

tom-pytel:
I don't think I've seen this mechanism in the other agents but I was asked to implement this. A cold start is when an instance is spun up to serve a request, for this reason the response time of the endpoint includes the spinup time of the instance, which is not ideal. This PR detects this condition and marks spans to which it applies.

Cold start detection is as follows: First span to run within 1 second of skywalking init is considered a cold start. This span gets the tag 'coldStart' set to 'true'. This span also optionally gets the text '\<cold\>' appended to the endpoint name if env var SW_COLD_ENDPOINT is set to 'true' or config option 'coldEndpoint', the default is off (normal endpoint names). A dummy span running first due to ignore is considered a valid first run and eats the cold start. Note: 'coldStart' tag is not added at all if cold start is not detected, which means most spans will not have anything extra added.

The metrics of the endpoint that was cold-started do not change (much) so this tag is not for the endpoint itself but for analysis of spans upstream which will be significantly slower due to the cold start. The visual part via 'coldEndpoint' is more for quick visual analysis in the trace view.

The target environment for this was Azure Functions which we are working on now but the mechanism is general so it should apply in any environment where instances are spun up on demand (Kubernetes). Performance-wise it is just a flag check so impact should be nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ Environment Variable | Description | Default
 | `SW_AGENT_SECURE` | Whether to use secure connection to backend OAP server | `false` |
 | `SW_AGENT_AUTHENTICATION` | The authentication token to verify that the agent is trusted by the backend OAP, as for how to configure the backend, refer to [the yaml](https://github.com/apache/skywalking/blob/4f0f39ffccdc9b41049903cc540b8904f7c9728e/oap-server/server-bootstrap/src/main/resources/application.yml#L155-L158). | not set |
 | `SW_AGENT_LOGGING_LEVEL` | The logging level, could be one of `error`, `warn`, `info`, `debug` | `info` |
-| `SW_AGENT_DISABLE_PLUGINS` | Comma-delimited list of plugins to disable in the plugins directory (e.g. "mysql", "express"). | `` |
+| `SW_AGENT_DISABLE_PLUGINS` | Comma-delimited list of plugins to disable in the plugins directory (e.g. "mysql", "express") | `` |
+| `SW_COLD_ENDPOINT` | Cold start detection is as follows: First span to run within 1 second of skywalking init is considered a cold start. This span gets the tag `coldStart` set to 'true'. This span also optionally gets the text '\<cold\>' appended to the endpoint name if SW_COLD_ENDPOINT is set to 'true'. | `false` |
 | `SW_IGNORE_SUFFIX` | The suffices of endpoints that will be ignored (not traced), comma separated | `.jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg` |
 | `SW_TRACE_IGNORE_PATH` | The paths of endpoints that will be ignored (not traced), comma separated | `` |
 | `SW_HTTP_IGNORE_METHOD` | Comma-delimited list of http methods to ignore (GET, POST, HEAD, OPTIONS, etc...) | `` |
--- a/src/Tag.ts
+++ b/src/Tag.ts
@@ -24,6 +24,7 @@ export interface Tag {
 }
 
 export default {
+  coldStartKey: 'coldStart',
   httpStatusCodeKey: 'http.status.code',  // TODO: maybe find a better place to put these?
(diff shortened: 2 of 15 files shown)


tom-pytel:
Forgot to update tests for coldStart tag.

README.md
```diff
@@ -56,7 +56,8 @@ Environment Variable | Description | Default
 | `SW_AGENT_SECURE` | Whether to use secure connection to backend OAP server | `false` |
 | `SW_AGENT_AUTHENTICATION` | The authentication token to verify that the agent is trusted by the backend OAP, as for how to configure the backend, refer to [the yaml](https://github.com/apache/skywalking/blob/4f0f39ffccdc9b41049903cc540b8904f7c9728e/oap-server/server-bootstrap/src/main/resources/application.yml#L155-L158). | not set |
 | `SW_AGENT_LOGGING_LEVEL` | The logging level, could be one of `error`, `warn`, `info`, `debug` | `info` |
-| `SW_AGENT_DISABLE_PLUGINS` | Comma-delimited list of plugins to disable in the plugins directory (e.g. "mysql", "express"). | `` |
+| `SW_AGENT_DISABLE_PLUGINS` | Comma-delimited list of plugins to disable in the plugins directory (e.g. "mysql", "express") | `` |
+| `SW_COLD_ENDPOINT` | When a cold start is detected append `<cold>` to the endpoint name | `false` |
```
kezhenxu94, line 60: I’d suggest put some of the following into the doc

> Cold start detection is as follows: First span to run within 1 second of skywalking init is considered a cold start. This span gets the tag 'coldStart' set to 'true'. This span also optionally gets the text '<cold>' appended to the endpoint name if env var SW_COLD_ENDPOINT is set to 'true' or config option 'coldEndpoint', the default is off (normal endpoint names). A dummy span running first due to ignore is considered a valid first run and eats the cold start. Note: 'coldStart' tag is not added at all if cold start is not detected, which means most spans will not have anything extra added.

kezhenxu94:
@wu-sheng WDYT of the cold start concept and mechanism?

wu-sheng:
In the code start case, is there any chance to separate the booting process and request handling process? These are typical differences in most cases I am familiar with. Or do you mean, this code starting begins after receiving a request? I don't know this kind of case.

tom-pytel:
> In the code start case, is there any chance to separate the booting process and request handling process? These are typical differences in most cases I am familiar with. Or do you mean, this code starting begins after receiving a request? I don't know this kind of case.

The `coldStart` tag doesn't really give information about the endpoint where it applies in isolation but rather explains the slower than normal execution of an upstream parent span which waits on this child. For one cold started child alone there is no way to determine the length of the bootup since the span starts once the boot is complete and in theory will execute more or less the same as if it was not a cold start (except for any unfilled caches and things like that). If you have both cold and non-cold versions of an endpoint then you can estimate the time of bootup alone by subtracting the normal non-coldStart average from the coldStart average, visually:
```
              0   1   2   3   4

/parent       +-------+
/child            +---+

/parent       +---------------+
/child<cold>      +-----------+

boot time = /child<cold> time - /child time = (4-1) - (2-1) = 2
```
In any case this is just something we do on our side because someone requested it, thought it might be useful info for your side as well.

wu-sheng:
I am good with a new endpoint with `<cold>` suffix. 
But this PR reminds me, I can't find the plugin dev doc about APIs explanation in NodeJS agent. Could you submit an issue to track this, and try to submit a PR to fix?

tom-pytel:
> I am good with a new endpoint with `<cold>` suffix.

That part is optional for visual inspection as well as separating statistics for warm vs. cold endpoints (or maybe help with some DB analysis queries as well). What is always added is the `coldStart` tag.

> But this PR reminds me, I can't find the plugin dev doc about APIs explanation in NodeJS agent. Could you submit an issue to track this, and try to submit a PR to fix?

This is more for @kezhenxu94 I think?

kezhenxu94 (approved):
LGTM

PR #64, opened 2021-11-12
The AxiosPlugin ignores SW8 if it is already in headers

nd-lqj:


wu-sheng:
We have told, we need all pull requests through English title and content.
Last time, PMC fixed for you, but would not be again.

nd-lqj:
> We have told, we need all pull requests through English title and content. Last time, PMC fixed for you, but would not be again.

Is that all right?

wu-sheng:
You should be more clear about what is the case. Do you mean multiple requests processed by one span?

nd-lqj:
> You should be more clear about what is the case. Do you mean multiple requests processed by one span?

Sw8 has already been handled on the client side, and the Node service only needs pass-through

wu-sheng:
I am confused, sw8 should never be forwarded.
sw8 is tracing context. The client side generates a new key/value for request, and server side decodes it.

nd-lqj:
> I am confused, sw8 should never be forwarded. sw8 is tracing context. The client side generates a new key/value for request, and server side decodes it.

The client uses AXIos to request the Node service, and the Node service uses AXIos to request the back-end service.  Sw8 is already created on the client side, so there is no need to create SW8 on the Node service side

src/plugins/AxiosPlugin.ts
```diff
@@ -45,11 +46,14 @@ class AxiosPlugin implements SwPlugin {
...
         config = url ? {...url} : {};
 
+      const carrier = ContextCarrier.from(config.headers || {});
       const {origin, host, pathname: operation} = new URL(config.url, config.baseURL);  // TODO: this may throw invalid URL
       const method = (config.method || 'GET').toUpperCase();
       const span = ignoreHttpMethodCheck(method)
         ? DummySpan.create()
-        : ContextManager.current.newExitSpan(operation, Component.AXIOS, Component.HTTP);
+        : carrier
+          ? ContextManager.current.newEntrySpan(operation, carrier)
```
tom-pytel, line 55: Why would a client be creating a server `EntrySpan`?

wu-sheng:
> The client uses AXIos to request the Node service, and the Node service uses AXIos to request the back-end service. Sw8 is already created on the client side, so there is no need to create SW8 on the Node service side

They are in different processes, or different context. This doesn't make sense.

kezhenxu94:
> > I am confused, sw8 should never be forwarded. sw8 is tracing context. The client side generates a new key/value for request, and server side decodes it.
> 
> 
> 
> The client uses AXIos to request the Node service, and the Node service uses AXIos to request the back-end service.  Sw8 is already created on the client side, so there is no need to create SW8 on the Node service side  

If your "Node service" serves as a proxy and you don't want that appear in topology/traces you should not install SkyWaling agent in it and pass all the headers in your Axios.

nd-lqj:
> > > I am confused, sw8 should never be forwarded. sw8 is tracing context. The client side generates a new key/value for request, and server side decodes it.
> > 
> > 
> > The client uses AXIos to request the Node service, and the Node service uses AXIos to request the back-end service.  Sw8 is already created on the client side, so there is no need to create SW8 on the Node service side
> 
> If your "Node service" serves as a proxy and you don't want that appear in topology/traces you should not install SkyWaling agent in it and pass all the headers in your Axios.

I think you are right.  My current solution is to disable the AxiosPlugin in the Node layer and have the HttpPlugin suffice.

src/plugins/AxiosPlugin.ts
nd-lqj: The AxiosPlugin also runs in the Node layer. So I think there's a distinction to be made here, that I'm not being captured in the right way, but the problem is there.

PR #65, opened 2021-11-26
Bump up gRPC version, and use its new release repository

alanlvle:
update:
1.grpc update to @grpc/grpc-js
2.update Authentication Method to realize


package.json
```diff
@@ -69,7 +70,6 @@
   },
   "dependencies": {
     "google-protobuf": "^3.14.0",
-    "grpc": "^1.10.1",
```
wu-sheng, line 72: No gRPC dependency?

scripts/protoc.sh
```diff
@@ -24,17 +24,16 @@ mkdir -p $OUT_DIR || true
 cd "${ROOT_DIR}"/protocol || exit
 
 PROTOC_GEN_TS_PATH="${ROOT_DIR}/node_modules/.bin/protoc-gen-ts"
-PROTOC_PLUGIN="${ROOT_DIR}/node_modules/.bin/grpc_tools_node_protoc_plugin"
+#PROTOC_PLUGIN="${ROOT_DIR}/node_modules/.bin/grpc_tools_node_protoc_plugin"
```
wu-sheng, line 27: Why comment out? Rather than deleting?

src/agent/protocol/grpc/AuthInterceptor.ts
```diff
@@ -17,23 +17,32 @@
...
+//       // tslint:disable-next-line:ban-types
+//       start(metadata: Metadata, listener: Listener, next: Function) {
+//         if (config.authorization) {
+//           metadata.add('Authentication', config.authorization);
+//         }
+//         next(metadata, listener);
+//       }
+//     })(),
+//   );
+// }
```
wu-sheng, line 40: Same question.

wu-sheng:
A lot of format changes, I can't see they are expected, please make sure you submitted the necessary changes.

package.json
alanlvle: use "@grpc/grpc-js": "^1.4.4"

wu-sheng: It is in `devDependencies` only

scripts/protoc.sh
alanlvle: new grpc not use grpc_tools_node_protoc_plugin.
https://www.npmjs.com/package/grpc_tools_node_protoc_ts

wu-sheng: This is not the point. The key is removing useless codes, not comment them out.

alanlvle: ok

package.json
alanlvle: ok

src/agent/protocol/grpc/clients/TraceReportClient.ts
```diff
@@ -53,14 +52,14 @@ export default class TraceReportClient implements Client {
...
-      emitter.emit('segments-sent');  // reset limiter in SpanContext
+      emitter.emit('segments-sent'); // reset limiter in SpanContext
 
       try {
         if (this.buffer.length === 0) {
           return;
         }
 
-        const stream = this.reporterClient.collect((error, _) => {
+        const stream = this.reporterClient.collect(AuthInterceptor(),(error, _) => {
```
kezhenxu94, line 62: Please add a space after `,`

src/agent/protocol/grpc/AuthInterceptor.ts
```diff
@@ -17,23 +17,14 @@
...
-        if (config.authorization) {
-          metadata.add('Authentication', config.authorization);
-        }
-        next(metadata, listener);
-      }
-    })(),
-  );
+export default function AuthInterceptor() { 
+  const mata = new grpc.Metadata()
+  if(config.authorization){
```
kezhenxu94, line 26: Please add a space before `(`, after `)`

alanlvle: update 3da6ed9155da8a0aebb36569956adae4bf382a02

kezhenxu94:
@alanlvle can you please also update the license information in this directory https://github.com/apache/skywalking-nodejs/tree/master/dist ? Others look good to me, thank you!

alanlvle:
@kezhenxu94 Yes, what are the changes?

wu-sheng:
Generally this part, https://github.com/apache/skywalking-nodejs/blob/master/dist/LICENSE#L211. You need to update the LICENSE file according to the latest dependencies, even dependencies of dependencies, which is required by the ASF and Apache 2.0 License.

kezhenxu94 (approved):

PR #69, opened 2022-02-16
fix move husky dependency to normal deps

tom-pytel:
Moved husky dependency from dev deps to normal deps, fails `npm install` from our scoped repository otherwise.


tkNobug:

kezhenxu94 (approved):

PR #70, opened 2022-02-21
experimental AWS Lambda Function support

tom-pytel:
Explicit AWS Lambda Function plugins in their own `aws/` subdirectory. Manual wrapping of AWS handler functions is required (much like the Azure Functions plugin), see the README. The following have been added:

* `AWSLambdaTriggerPlugin` a generic wrapper which should work with all types of function triggers, but at the cost of information and linking back to calling endpoint.
* `AWSLambdaGatewayAPIHTTP` specific wrapper for GatewayAPI HTTP type triggers, stores HTTP info and status code and allows linking back to calling endpoint.
* `AWSLambdaGatewayAPIREST` same as above except for the GatewayAPI REST type trigger.
* `agent.flush()` mechanism for synchronizing on segment data send. This is needed for AWS Lambdas because AWS freezes processes in between invocations.
* `ContextManager.removeTailFinishedContexts()` to remove any finished spans from a context to break a parent-child chain which is invalid because sequential independent invocations of a handler are executed in inherited async contexts of previous runs instead of being executed hierarchically like most servers do (for AWS Lambda functions).

I will send up the component IDs to the main `skywalking` repo once tests are passed.

### DIFF
--- a/src/trace/Component.ts
+++ b/src/trace/Component.ts
@@ -28,6 +28,9 @@ export class Component {
   static readonly RABBITMQ_PRODUCER = new Component(52);
   static readonly RABBITMQ_CONSUMER = new Component(53);
   static readonly AZURE_HTTPTRIGGER = new Component(111);
+  static readonly AWSLAMBDA_FUNCTION = new Component(124);
+  static readonly AWSLAMBDA_GATEWAYAPIHTTP = new Component(125);
+  static readonly AWSLAMBDA_GATEWAYAPIREST = new Component(126);
   static readonly EXPRESS = new Component(4002);
   static readonly AXIOS = new Component(4005);
   static readonly MONGOOSE = new Component(4006);
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,5 +17,8 @@ module.exports = {
   },
   env: {
     node: true
+  },
+  globals: {
+    "NodeJS": true
   }
 };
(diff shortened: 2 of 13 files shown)


src/trace/Component.ts
```diff
@@ -28,6 +28,9 @@ export class Component {
   static readonly RABBITMQ_PRODUCER = new Component(52);
   static readonly RABBITMQ_CONSUMER = new Component(53);
   static readonly AZURE_HTTPTRIGGER = new Component(111);
+  static readonly AWSLAMBDA_FUNCTION = new Component(124);
+  static readonly AWSLAMBDA_GATEWAYAPIHTTP = new Component(125);
+  static readonly AWSLAMBDA_GATEWAYAPIREST = new Component(126);
```
kezhenxu94, line 33: Can you find some proper IDs for these plugins?

https://github.com/apache/skywalking/blob/68e40eb9fc0a891a55fe927d49b267d0fd80000c/oap-server/server-starter/src/main/resources/component-libraries.yml#L459

kezhenxu94 (commented):
And please check the CI failures

src/trace/Component.ts
tom-pytel: Yes they will be fixed, it may be a little bit though as I have several things at the moment. The IDs are already found, just haven't sent them up to the main SW repo, will do when this is fixed.

kezhenxu94: > Yes they will be fixed, it may be a little bit though as I have several things at the moment. The IDs are already found, just haven't sent them up to the main SW repo, will do when this is fixed.

These IDs 124~126 seems to be in the reserved range for other language, not NodeJS, so better choose those inside the range for NodeJS, (e.g. `400` ~ `5000`)

https://github.com/apache/skywalking/blob/68e40eb9fc0a891a55fe927d49b267d0fd80000c/oap-server/server-starter/src/main/resources/component-libraries.yml#L459-L460

tom-pytel: That is the same place we put the Azure Functions ID. Also, these are not limited to NodeJS, you can run lots of languages on Lambda with these triggers.

tom-pytel: Ok, so do I push the IDs I have to main repo or do you want change?

kezhenxu94: > Ok, so do I push the IDs I have to main repo or do you want change?

Please do that, using 124~126 is good to me

kezhenxu94 (approved):

tom-pytel:
Will merge when main sw repo accepts component IDs.

PR #73, opened 2022-02-25
Update IORedis Plugin, fill dbinstance tag as host if condition.select doesn't exist.

TonyKingdom:
fix issue: https://github.com/apache/skywalking/issues/7275

### DIFF
--- a/src/plugins/IORedisPlugin.ts
+++ b/src/plugins/IORedisPlugin.ts
@@ -49,7 +49,7 @@ class IORedisPlugin implements SwPlugin {
       span.layer = SpanLayer.CACHE;
       span.peer = host;
       span.tag(Tag.dbType('Redis'));
-      span.tag(Tag.dbInstance(`${this.condition?.select??host}`));
+      span.tag(Tag.dbInstance(`${this.condition?.select ?? host}`));
 
       try {
         const ret = wrapPromise(span, _original.apply(this, args));
--- a/src/plugins/IORedisPlugin.ts
+++ b/src/plugins/IORedisPlugin.ts
@@ -49,7 +49,7 @@ class IORedisPlugin implements SwPlugin {
       span.layer = SpanLayer.CACHE;
       span.peer = host;
       span.tag(Tag.dbType('Redis'));
-      span.tag(Tag.dbInstance(`${this.condition.select}`));
+      span.tag(Tag.dbInstance(`${this.condition?.select}`));
 
       try {
         const ret = wrapPromise(span, _original.apply(this, args));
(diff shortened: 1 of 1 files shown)


src/plugins/IORedisPlugin.ts
```diff
@@ -49,7 +49,7 @@ class IORedisPlugin implements SwPlugin {
       span.layer = SpanLayer.CACHE;
       span.peer = host;
       span.tag(Tag.dbType('Redis'));
-      span.tag(Tag.dbInstance(`${this.condition.select}`));
+      span.tag(Tag.dbInstance(`${this.condition?.select}`));
```
wu-sheng, line 52: Could the `condition` be null? Could you share in which cases?

TonyKingdom: As described in https://github.com/apache/skywalking/issues/7275,  here is what happened in my app:
image

wu-sheng: I got this. I mean is there a solution to make sure `dbInstance` would be set properly? Because this is an important tag, it is expected by the backend.

TonyKingdom: As we use redis cluster, `dbInstance` makes no sense. Also, there's no `condition` property in redis cluster mode.

wu-sheng: For cluster mode, instance should be a cluster name or IP port list.
This is recommended.

TonyKingdom: Redis cluster mode can be  inferred by span,  and here `dbInstance` actually means zero-based numeric index of Redis accounding to the repo.

image

wu-sheng: There is no cache dashboard to represent metrics for performance yet, but once it is added, this undefined name will show up in the cache server list. 
So, only an endpoint name is not enough. I can see peer includes an IP:port. Isn't it for the Redis server?

TonyKingdom: The Peer here stands for one of Redis cluster nodes. Is this value suitable for `dbInstance `?

wu-sheng: I am expecting the whole `IP1:port1, IP2:port2,...`

TonyKingdom: Redis cluster nodes cannot be acquired by the code contexts unless we call some function which I believe is not appropriate.

wu-sheng: Then at least use the peer value. It is at least better than null.

wu-sheng:
Maybe not perfect, but good for now.

@kezhenxu94 WDYT?

kezhenxu94 (approved):
Good to me

PR #74, opened 2022-03-10
Fix mysql2 plugin install error

ruleeeer:
I've added an optional method to get the version information of the plugin. For most plugins it's not necessary to implement this method, the default `require('lib/package.json')` method can get its version, but the few libraries that specify `export` in `package.json` can't use default method. They must implement their own `getVersion()` method to get the version

Fix https://github.com/apache/skywalking/issues/7879

### DIFF
--- a/src/core/OptionMethods.ts
+++ b/src/core/OptionMethods.ts
@@ -1,0 +1,24 @@
+/*!
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


ruleeeer:
Also, for `require('mysql/lib/connect')` change to `require('mysql').Connection` also because of the restriction of the export field,The meaning of the export field can be viewed here:https://nodejs.org/dist/latest-v16.x/docs/api/packages.html
image

The `export` field of mysql2 is as follows:
``` json
"exports": {
    ".": "./index.js",
    "./promise": "./promise.js",
    "./promise.js": "./promise.js"
  }
```
This also means that except for require('mysql2'), require('mysql2/promise') and require('mysql2/promise.js'), all other requires are illegal, such as require('mysql2/lib/connection' ) and require('mysql2/package.json')

ruleeeer:
@wu-sheng 
Could help to do this pr associated to Issue:https://github.com/apache/skywalking/issues/7879 ,thanks

wu-sheng:
image

This should work.

kezhenxu94 (approved):
Thank you @ruleeeer

PR #80, opened 2022-04-19
support mongodb@4

bangbang93:


package.json
```diff
@@ -56,12 +56,12 @@
...
     "eslint-plugin-prettier": "^4.0.0",
     "express": "^4.17.1",
-    "grpc-tools": "^1.11.1",
     "grpc_tools_node_protoc_ts": "^5.3.0",
+    "grpc-tools": "^1.11.1",
     "ioredis": "^4.27.2",
     "jest": "^26.6.3",
     "lint-staged": "^12.3.2",
-    "mongodb": "^3.6.4",
+    "mongodb": "^4.5.0",
```
wu-sheng, line 64: I am not sure whether this is a good practise.

@tom-pytel @kezhenxu94 How nodeJS agent controls the versions of plugin supporting?

bangbang93: it's in the devDependencies, take not effect to the application

kezhenxu94 (dismissed):

wu-sheng:
It seems CI fails. May need some recheck.

wu-sheng:
10k lines changed? I don't see this coming. Please verify.

bangbang93:
> 10k lines changed? I don't see this coming. Please verify.

it's `package-lock.json`, I'm using node@17, which bundles npm@8.5.2, the lock file version is different.

wu-sheng:
We should not update all. And you need to follow the upstream lock version. Otherwise, different people will change the files from time to time.

bangbang93:
But I need to update the mongodb dev dependence, this file must be changed.

wu-sheng:
Changing is fine, but should use the same npm version it was. Then the changes are minimum.

kezhenxu94:
@bangbang93 do you want to continue this?

bangbang93:
Works busy. I will update in a few days.

wu-sheng:
@bangbang93 According to this, https://github.com/apache/skywalking/discussions/9070#discussioncomment-2786199, this change works.
Could you make CI passed?

@kezhenxu94 I am not sure how NodeJS agent works/compiling/running with different MongoDB client versions? I can see this change upgrades the 3.x to 4.x. I think we should have a way to work for both?

bangbang93:
I have no idea why the ci keeps failing. I'm using my own fork with the same change in my production, and it works well too.

kezhenxu94:
@bangbang93 I think there is some minor different behaviors between mongodb@3 and mongodb@4, see the test case, there is no `MongoDB/collection` endpoint in mongodb@4 , which exists in mongodb@3

https://github.com/apache/skywalking-nodejs/runs/6517919227?check_suite_focus=true#step:5:13013

warrmr:
Updating the Dev dependancies in the package.json appears to have caused the test failure due to an update in the MongoDB API between version 3 and 4. 

https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md#dbcollection-no-longer-accepts-a-callback

Either the dev dependancy in package.json will need to be rolled back or the test at the following line will need to be refactored. 

https://github.com/apache/skywalking-nodejs/blob/92fe2adac6a14ceb5b3c4881010c59b82e064e56/tests/plugins/mongodb/server.ts#L38


Could be refactored like this and the test should then pass with the updated dev dependancy. 

```JavaScript
try {
    const resDB = await client.db('admin').collection('docs').findOne();
    res.end(`${resDB}`);
    resolve(null);
    client.close();
} catch (err: any) {
    res.end(`${err}`);
    resolve(null);
    client.close();
}
```

However I would pose the question is there a way that we can run the test suite for both versions of MondoDB client the previous version in the package.json and the updated version. Or is updating the test to match the updated API an acceptable solution?

kezhenxu94:
> However I would pose the question is there a way that we can run the test suite for both versions of MondoDB client the previous version in the package.json and the updated version? Or is updating the test to match the updated API an acceptable solution?

We could do some extra adjustments to the test code to run tests on different versions but that can be another task (PR), but for now I prefer we refactor the codes to make it works for both 3 and 4

warrmr:
So I got bored and decided to have a play. I have forked off and have got @bangbang93 's changes working on my fork without making 182 changes to the package-lock file and it all passes good in a test PR on my fork.

https://github.com/warrmr/skywalking-nodejs/pull/1


This is very much TMIW (Trust Me It Works) as the tests are testing Mongo@3.6 so is not testing the new changes but ensures the old version still works. 

I have also added a .nvmrc file on my fork to specify the node version as I had the same problem with the lock-file getting updated. locking to Node version 12 on my dev machine kept the lock file in a good shape. 


Next steps would be to install MongoDB@4 in the dev-dependancy and write an additional test/test file for each version of MongoDB client and then we have tested the new code and ensure both work both ways. 


@kezhenxu94 what do you think to the update in my PR on my fork, is it worth me trying to write some tests so it is tested on both 3 and 4 versions. as currently only testing on 3 version. ? 

@bangbang93  or do you want to fix CI and I will leave it for you?

kezhenxu94:
> So I got bored and decided to have a play. I have forked off and have got @bangbang93 's changes working on my fork without making 182 changes to the package-lock file and it all passes good in a test PR on my fork.
> 
> 
> 
> https://github.com/warrmr/skywalking-nodejs/pull/1
> 
> 
> 
> 
> 
> This is very much TMIW (Trust Me It Works) as the tests are testing Mongo@3.6 so is not testing the new changes but ensures the old version still works. 
> 
> 
> 
> I have also added a .nvmrc file on my fork to specify the node version as I had the same problem with the lock-file getting updated. locking to Node version 12 on my dev machine kept the lock file in a good shape. 
> 

Hi, it'd be nice if you could submit the nvmrc in a separate PR so future developers won't get the same problem in terms of package lock file. 
> 
> 
> 
> Next steps would be to install MongoDB@4 in the dev-dependancy and write an additional test/test file for each version of MongoDB client and then we have tested the new code and ensure both work both ways. 
> 
> 
> 
> 
> 
> @kezhenxu94 what do you think to the update in my PR on my fork, is it worth me trying to write some tests so it is tested on both 3 and 4 versions. as currently only testing on 3 version. ? 
> 

Sure, sounds good to me. That would be awesome!

wu-sheng:
This should be closed if no further update.

PR #81, opened 2022-05-11
improved ignore path regex

tom-pytel:
Minor enhancement/fix? Previously the config.traceIgnorePath pattern `/a/**/*` would match `/a/b/`, `/a/b/c` and `/a/b/c/`, etc... but not `/a/` or `/a/b`. Now it does.


tkNobug:

kezhenxu94 (approved):

PR #82, opened 2022-05-19
Fix skywalking s3 upload error

wuwen5:
- [x] fix https://github.com/apache/skywalking/issues/8824


kezhenxu94 (approved):
Tested and it works, thanks @wuwen5 👍

PR #84, opened 2022-06-06
Fix issue #9165 express router entry duplicated

tom-pytel:
See said issue: https://github.com/apache/skywalking/issues/9165#event-6729419470


tom-pytel:
Have tested with our fork and works. Someone should test vanilla. 

P.S. The linter being used by this project currently is pure evil incarnate. If evil could be distilled into software it is this linter.

kezhenxu94 (approved):

PR #90, opened 2022-07-16
Fix sw header is not validated

kezhenxu94:
@ypg521 will you be able to help to test this fix?

Closes https://github.com/apache/skywalking/issues/7505


kezhenxu94:
Once verified, I'll start to release 0.5.1

wu-sheng (approved):

ypg521:
Verified , it's right 💯

PR #96, opened 2022-10-14
fix: entry span url in endponts using Express middleware/router objects

BFergerson:
The following code:
```typescript
const app = express();
const testRouter = express.Router();

app.use('/test', testRouter);
testRouter.get('/express', (req, res) => {
    res.sendStatus(200);
});

app.listen(3000);
```

Results in entry spans with the URL of `/express` and not the correct URL of `/test/express`.

This PR solves that using [req.originalUrl](https://expressjs.com/en/4x/api.html#req.originalUrl) when available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.5.2
+- Correct entry span url in endponts using Express middleware/router objects
+
 # 0.5.1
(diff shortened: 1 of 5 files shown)


kezhenxu94 (approved):
Nice catch!

PR #98, opened 2022-10-14
chore: use openapi format for endpoint uris

BFergerson:
fixes https://github.com/apache/skywalking/issues/9787


kezhenxu94 (approved):

kezhenxu94:
Is it still WIP?

BFergerson:
I thought it would be but ended up fixing it on the first try. This is good to go.

PR #100, opened 2022-11-12
AWS DynamoDB, Lambda, SQS and SNS plugins, webpack

tom-pytel:
The main things here are 4 new AWS SDK2 plugins and webpack support. The full list is:

* AWS SDK2 DynamoDB plugin.
* AWS SDK2 Lambda plugin.
* AWS SDK2 SQS plugin.
* AWS SDK2 SNS plugin.
* Instrument deprecated but possibly still used exit calls `done()`, `succeed()` and `fail()` from AWS Lambda function.
* Allow passing trace ID in AWS Lambda function parameters to allow linking outgoing to incoming calls.
* Extend AWS flush from boolean always/never to time threshold.
* Made agent.flush() wait on any unfinished Spans before waiting for protocol send to make sure all Spans in progress finish.
* Webpack support, can now run with most plugins from a bundle where before agent did not work at all from a bundle.
* Changed builtin plugin "detection" to explicit specification by plugin.
* Fix sequential invocation of Lambda function which does not flush and leaves an async component waiting to complete a Span incorrectly linking to a previous sibling Span.
* Fix double logging and undefined version print if plugin not supported.
* Fix Axios plugin to not return error but rather empty not supported version string if not supported (is deprecated anyway due to not compatible with Axios v1.0+, need to update).

The AWS plugins are in the `plugins/` directory and not the `aws/` directory because they are normal Node plugins which can be used on or off AWS to access those AWS services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,9 @@ Environment Variable | Description | Default
 | `SW_SQL_PARAMETERS_MAX_LENGTH` | The maximum string length of SQL parameters to log | `512` |
 | `SW_MONGO_TRACE_PARAMETERS` | If set to 'true' then mongodb query parameters will be included | `false` |
 | `SW_MONGO_PARAMETERS_MAX_LENGTH` | The maximum string length of mongodb parameters to log | `512` |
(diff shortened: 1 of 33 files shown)


tom-pytel:
Good to go.

kezhenxu94 (approved):

PR #102, opened 2022-11-14
Update AxiosPlugin for v1.0+

tom-pytel:
Old plugin doesn't work with v1.0+, dropping support for < v1.0. Tests will probably break so will see how to fix.


tom-pytel:
Newest Axios seems to be incompatible with Node 10. Suggest dropping Node 10 support since lowest Node currently getting security updates is 14 and latest version is 19.

wu-sheng:
v1.0 is released one month ago, https://github.com/axios/axios/releases/tag/v1.0.0

Are we too aggressive to drop the old version release? @kezhenxu94 Do we have a way to build plugin for various versions? As a result, we could keep the plugin for 0.x and one more for 1.x

tom-pytel:
> Are we too aggressive to drop the old version release? @kezhenxu94 Do we have a way to build plugin for various versions? As a result, we could keep the plugin for 0.x and one more for 1.x

Not a big problem to leave it compatible with old Axios for old Node 10, this should work for all versions, lets see what tests say. I left test version as old Axios because otherwise Node 10 test would fall over and die.

src/plugins/AxiosPlugin.ts
```diff
@@ -26,20 +26,39 @@ import { SpanLayer } from '../proto/language-agent/Tracing_pb';
...
+  getVersion(installer: PluginInstaller): string {
+    // TODO: this method will not work in a bundle
+    try {
+      const indexPath = installer.resolve(this.module);
+      const dirname = indexPath.slice(
+        0,
+        indexPath.lastIndexOf(`${path.sep}node_modules${path.sep}axios${path.sep}`) + 20,
+      );
+      const packageSJonStr = fs.readFileSync(`${dirname}package.json`, { encoding: 'utf-8' });
+      const pkg = JSON.parse(packageSJonStr);
```
kezhenxu94, line 45:
```suggestion
      const packageJsonStr = fs.readFileSync(`${dirname}package.json`, { encoding: 'utf-8' });
      const pkg = JSON.parse(packageJsonStr);
```

tom-pytel: 🥇 :|

kezhenxu94 (approved):

PR #107, opened 2022-12-19
fix: add namespace for trace id prefix

wuliupo:


wu-sheng:
`sw8` is the required key in the API. This PR can't be accepted.

tkNobug:

wu-sheng:
I believe we had removed that from Java agent long time ago. Maybe you looked at old blogs?

wu-sheng:
That was a long time ago decision as it is easy to break other integration. Not every place adopted this, and it's not widely used.

zhaoxiaojie0415:
> That was a long time ago decision as it is easy to break other integration. Not every place adopted this, and it's not widely used.

Got it, thanks

PR #118, opened 2023-09-03
feat: add deadline for trace request

ErosZy:
feat: add deadline for trace request

In our practice, we have found that when skywalking's backend service is under too much pressure to handle the request quickly, the trace request will stay alive by default and not be closed (the default value of grpc deadline is [2147483647](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/deadline.ts#L60C26-L60C36)). If the default `maxBufferSize` is used, a large number of requests per second will pile up, and slowly the client's memory will rise until it reaches the pod's limit and restarts. as shown:

截屏2023-09-03 下午9 47 09

This also causes frequent up and down for skywalking's backend services (because of RetryingCall), which is very dangerous!

265268504-ed146317-35a2-4399-93bc-aaef7aebabc4


tkNobug:

kezhenxu94 (approved):

PR #119, opened 2023-10-19
Setup node 18 for testing matrix

wu-sheng:


tkNobug:

wu-sheng:
@kezhenxu94 Should we add the 18 here? 

BTW, https://github.com/apache/skywalking-nodejs/pull/118 has not been released. Although we have that after 0.6, should we consider a new one?

kezhenxu94 (approved):

PR #123, opened 2024-06-20
fix tag name

heyanlong:
Fix the OAL `httpResponseStatusCode` problems 
Because the `OAP` service uses `http.status_code` as analysis field, we need to fix it

https://github.com/apache/skywalking/blob/2603d2bd17a09723ebef0c5819f7c303b776f6d1/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/RPCAnalysisListener.java#L222

### DIFF
--- a/src/Tag.ts
+++ b/src/Tag.ts
@@ -25,7 +25,7 @@ export interface Tag {
 
 export default {
   coldStartKey: 'coldStart',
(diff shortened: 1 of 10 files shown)


tkNobug:

wu-sheng (approved):

PR #129, opened 2026-01-28
fix: add destroy methods to prevent memory leaks in protocol clients

wolfsilver:
In the event of prolonged network outages, this may lead to severe memory issues.

This pull request introduces a resource cleanup mechanism across the agent protocol and its gRPC client implementations to prevent memory leaks and ensure graceful shutdown. The main changes add a standardized optional `destroy()` method to key interfaces and classes, and implement cleanup logic for timers, event listeners, and buffers.

Resource cleanup and lifecycle management:

* Added an optional `destroy()` method to the `Protocol` and `Client` interfaces (`Protocol.ts`, `Client.ts`), and implemented it in `GrpcProtocol`, `HeartbeatClient`, and `TraceReportClient` to clean up resources such as timers, event listeners, and buffers. 

Memory leak prevention in gRPC clients:

* In `HeartbeatClient`, implemented `destroy()` to clear the heartbeat timer and log cleanup.
* In `TraceReportClient`, refactored event listener registration to store a reference for proper removal, and implemented `destroy()` to remove the event listener, clear timeouts and buffers, and log cleanup. Also added logic to limit buffer size and discard oldest segments with a warning to prevent memory leaks during network issues.

Agent shutdown support:

* Added a `destroy()` method to the `Agent` class to call the protocol's `destroy()` method, reset protocol state, and log the shutdown process.

### DIFF
--- a/src/agent/protocol/Protocol.ts
+++ b/src/agent/protocol/Protocol.ts
@@ -28,4 +28,6 @@ export default interface Protocol {
   report(): this;
 
   flush(): Promise<any> | null;
+
+  destroy?(): void;
 }
--- a/src/agent/protocol/grpc/GrpcProtocol.ts
+++ b/src/agent/protocol/grpc/GrpcProtocol.ts
@@ -47,4 +47,10 @@ export default class GrpcProtocol implements Protocol {
   flush(): Promise<any> | null {
     return this.traceReportClient.flush();
   }
+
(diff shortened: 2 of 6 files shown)


a526672351:
已收到！

tkNobug:

kezhenxu94 (approved):

PR #139, opened 2026-06-27
Refactor gRPC runtime services and add Node.js runtime metrics

songzhendong:
> **Branch:** `feature/nodejs-grpc-v1-runtime-metrics` (based on upstream `master` @ `1d3ba73`)  
> **Commit:** `481b290` (`481b290beadfc530e4b0adef05dbbbb211af0a66`) — single commit, 26 files changed

### CI status (fork validation, commit `481b290`)

All checks **passed** on fork before upstream submission (2026-06-26, 34/34):

| Workflow | Result | Run |
|----------|--------|-----|
| **Build** (Node 20 / 22 / 24) | ✅ | https://github.com/songzhendong/skywalking-nodejs/actions/runs/28235029528 |
| **Test** (Lint / TestLib / TestPlugins) | ✅ | https://github.com/songzhendong/skywalking-nodejs/actions/runs/28235029569 |
| **License** | ✅ | https://github.com/songzhendong/skywalking-nodejs/actions/runs/28235029547 |

Key job (runtime meter E2E):

- **TestPlugins (20, express)** — `meterItems` asserts server + client × 6 meters each: https://github.com/songzhendong/skywalking-nodejs/actions/runs/28235029569/job/83647759105

Fork PR checks: https://github.com/songzhendong/skywalking-nodejs/pull/2/checks

### OAP dependency (not in this PR)

Runtime meters are reported via `MeterReportService`, but **OAP consumption requires a separate upstream PR** adding the `nodejs-runtime` analyzer (`nodejs-runtime.yaml`, 6 MAL rules). Agent-side reporting and plugin E2E `meterItems` are covered here.

---

### Summary

This PR does two things:

1. **gRPC architecture v1** — Refactor Node remote reporting to match Java **`agent.core.*`**: single Channel, unified `BootService` lifecycle, Builder/Decorator chain; Trace / Heartbeat / Meter share one code path for cross-language alignment.
2. **Runtime metrics** — Six `instance_nodejs_*` process meters via `MeterReportService` (1s collect / 1s report).

**Not in this PR:** release scripts, version bump, `CONTRIBUTING` / How-to-release, httpbin test infra, OAP analyzer.

### Motivation

Before refactor, Node used a standalone `GrpcProtocol` with a different structure from Java `agent.core`. This PR converges remote reporting onto the same architecture as Java: one shared gRPC channel, unified lifecycle and extension points, lower cross-language maintenance cost, and a path consistent with Java for failover, TLS, Event/Log, etc.

[...]

PR #143, opened 2026-07-07
feat(runtime): add six instance_nodejs_* meters (6 -> 12)

songzhendong:
Extend Node.js runtime metrics from **6 to 12** `instance_nodejs_*` meters reported via `MeterReportService.collect`, building on the six meters introduced in #139.

This PR is **metrics-only** — no gRPC/TLS/DNS/CommandService changes.

### Motivation

 Operators need richer Node.js process/V8 visibility (array buffers, uptime, malloc stats, heap generations) alongside the existing CPU/heap/RSS meters.

#### Six additional meters

| Node.js source | Meter name | Unit |
| :--- | :--- | :--- |
| `process.memoryUsage().arrayBuffers` | `instance_nodejs_array_buffers` | bytes |
| `process.uptime()` | `instance_nodejs_uptime` | seconds |
| `v8.getHeapStatistics().peak_malloced_memory` | `instance_nodejs_peak_malloced_memory` | bytes |
| `v8.getHeapStatistics().malloced_memory` | `instance_nodejs_malloced_memory` | bytes |
| `v8.getHeapSpaceStatistics()` old_space | `instance_nodejs_old_space_used` | bytes |
| `v8.getHeapSpaceStatistics()` new_space | `instance_nodejs_new_space_used` | bytes |

**12 total** with the six baseline meters: `process_cpu`, `heap_used`, `heap_total`, `heap_limit`, `rss`, `external_memory`.

#### Other

- `SW_AGENT_RUNTIME_METRICS_HEAP_SPACE_DETAIL` (default `true`); when `false`, skip `getHeapSpaceStatistics()` and report `old_space_used` / `new_space_used` as `0`.
- `SW_AGENT_RUNTIME_METRICS_MAX_SNAPSHOTS_PER_REPORT` (default `1`); drain at most one buffered snapshot per report tick (limits meter stream burst after reconnect).
- `MeterSender`: use per-snapshot `collectedAt` for `setTimestamp()` instead of a single batch `Date.now()`.
- Process CPU: `process.cpuUsage()` user + system, normalized by **logical CPU count** (0–100%).
- README: twelve-meter table + env documentation.
- Unit tests: `tests/runtime/RuntimeSampler.test.ts`, `tests/runtime/RuntimeMetricsCollector.test.ts`, `tests/remote/MeterSender.test.ts`.
- CI: `TestUnitRemote` matrix on Node **20 / 22 / 24** (`tests/remote`, `tests/config`, `tests/runtime`).
- Express plugin e2e: `meterSize: 12`, assertions for all twelve meters.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -73,7 +73,7 @@ jobs:
       - name: Unit tests on Node@${{ matrix.node-version }}
         run: |
           npm i
-          npx jest --testPathIgnorePatterns "/tests/plugins/" --runInBand
+          npx jest --testPathIgnorePatterns "/node_modules/" "/tests/plugins/" --runInBand
 
   build-matrix:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,7 +47,7 @@ jobs:
           npm i
           npm run lint
 
-  TestUnitRemote:
+  TestUnit:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     strategy:
@@ -73,7 +73,7 @@ jobs:
       - name: Unit tests on Node@${{ matrix.node-version }}
         run: |
           npm i
-          npx jest tests/remote/ tests/config/ tests/runtime/ --runInBand
+          npx jest --testPathIgnorePatterns "/tests/plugins/" --runInBand
 
   build-matrix:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,6 +47,34 @@ jobs:
           npm i
           npm run lint
 
+  TestUnitRemote:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        node-version: [ 20, 22, 24 ]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: Set Up NodeJS ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Unit tests on Node@${{ matrix.node-version }}
+        run: |
+          npm i
+          npx jest tests/remote/ tests/config/ tests/runtime/ --runInBand
+
   build-matrix:
     runs-on: ubuntu-latest
     timeout-minutes: 5
--- a/tests/plugins/express/client.ts
+++ b/tests/plugins/express/client.ts
@@ -37,12 +37,12 @@ app.use('/test', testRouter);
 
 testRouter.get('/express', (req, res) => {
   http
-    .request(`http://${process.env.SERVER || 'localhost:5000'}${req.url}`, (r) => {
-      let data = '';
-      r.on('data', (chunk) => (data += chunk));
-      r.on('end', () => res.send(data));
-    })
-    .end();
+  .request(`http://${process.env.SERVER || 'localhost:5000'}${req.url}`, (r) => {
+    let data = '';
+    r.on('data', (chunk) => (data += chunk));
+    r.on('end', () => res.send(data));
+  })
+  .end();
 });
 
 app.listen(5001, () => console.info('Listening on port 5001...'));
--- a/tests/plugins/express/client.ts
+++ b/tests/plugins/express/client.ts
@@ -25,9 +25,7 @@ agent.start({
   serviceName: 'client',
   maxBufferSize: 1000,
   // Plugin meter assertions need frequent reports within the test window.
-  runtimeMetricsCollectPeriod: 1000,
   runtimeMetricsReportPeriod: 1000,
-  runtimeMetricsUptimeReportPeriod: 1000,
 });
 
 const app = express();
--- a/tests/plugins/express/client.ts
+++ b/tests/plugins/express/client.ts
@@ -24,6 +24,10 @@ import express from 'express';
 agent.start({
   serviceName: 'client',
   maxBufferSize: 1000,
+  // Plugin meter assertions need frequent reports within the test window.
+  runtimeMetricsCollectPeriod: 1000,
+  runtimeMetricsReportPeriod: 1000,
+  runtimeMetricsUptimeReportPeriod: 1000,
 });
 
 const app = express();
@@ -33,12 +37,12 @@ app.use('/test', testRouter);
 
 testRouter.get('/express', (req, res) => {
   http
-  .request(`http://${process.env.SERVER || 'localhost:5000'}${req.url}`, (r) => {
-    let data = '';
-    r.on('data', (chunk) => (data += chunk));
-    r.on('end', () => res.send(data));
-  })
-  .end();
+    .request(`http://${process.env.SERVER || 'localhost:5000'}${req.url}`, (r) => {
+      let data = '';
+      r.on('data', (chunk) => (data += chunk));
+      r.on('end', () => res.send(data));
+    })
+    .end();
 });
 
 app.listen(5001, () => console.info('Listening on port 5001...'));
--- a/README.md
+++ b/README.md
@@ -72,19 +72,17 @@ Environment Variable | Description | Default
 | `SW_AWS_SQS_CHECK_BODY` | Incoming SQS messages check inside the body for trace ID in order to allow linking outgoing SNS messages to incoming SQS. | `false` |
 | `SW_AGENT_MAX_BUFFER_SIZE` | The maximum buffer size before sending the segment data to backend | `'1000'` |
 | `SW_AGENT_TRACE_TIMEOUT` | The timeout for trace requests to backend services | `'10000'` |
-| `SW_AGENT_NODEJS_RUNTIME_METRICS_REPORTER_ACTIVE` | Whether to report Node.js runtime metrics through MeterReportService (default: collect 20s, report 20s) | `true` |
-| `SW_AGENT_NODEJS_RUNTIME_METRICS_COLLECT_PERIOD` | Runtime metric sample interval in milliseconds | `20000` |
-| `SW_AGENT_NODEJS_RUNTIME_METRICS_REPORT_PERIOD` | Runtime metric report interval in milliseconds (aligned with Java `meter.report_interval`) | `20000` |
-| `SW_AGENT_NODEJS_RUNTIME_METRICS_UPTIME_REPORT_PERIOD` | How often to include `instance_nodejs_uptime` in a report (ms); other runtime meters still follow the report period | `30000` |
+| `SW_AGENT_NODEJS_RUNTIME_METRICS_REPORTER_ACTIVE` | Whether to report Node.js runtime metrics through MeterReportService (default period 20s) | `true` |
+| `SW_AGENT_NODEJS_RUNTIME_METRICS_REPORT_PERIOD` | Runtime metric sample + report interval in milliseconds (aligned with Java `meter.report_interval`) | `20000` |
 
-Legacy env names `SW_AGENT_RUNTIME_METRICS_*` / `SW_AGENT_NVM_*` for reporter active, collect period, and report period are still accepted as deprecated aliases.
+Legacy env names `SW_AGENT_RUNTIME_METRICS_*` / `SW_AGENT_NVM_*` for reporter active and report period are still accepted as deprecated aliases.
 
 
 Note that the various ignore options like `SW_IGNORE_SUFFIX`, `SW_TRACE_IGNORE_PATH` and `SW_HTTP_IGNORE_METHOD` as well as endpoints which are not recorded due to exceeding `SW_AGENT_MAX_BUFFER_SIZE` all propagate their ignored status downstream to any other endpoints they may call. If that endpoint is running the Node Skywalking agent then regardless of its ignore settings it will not be recorded since its upstream parent was not recorded. This allows the elimination of entire trees of endpoints you are not interested in as well as eliminating partial traces if a span in the chain is ignored but calls out to other endpoints which are recorded as children of ROOT instead of the actual parent.
 
 ## Node.js Runtime Metrics
 
-The agent reports twelve process-level meters (`instance_nodejs_*`) via `MeterReportService` by default (collect 20s, report 20s; `instance_nodejs_uptime` every 30s). Set `SW_AGENT_NODEJS_RUNTIME_METRICS_REPORTER_ACTIVE=false` to disable. Process CPU combines `process.cpuUsage()` user + system, normalized by logical CPU count (0–100%).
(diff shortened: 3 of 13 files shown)


wu-sheng (dismissed):
Thanks for the follow-up on #139. The six new meters themselves are straightforward and the `process.cpuUsage()` fix is a real improvement. Most of my comments are about the `MeterSender` rework that rode along with them.

## The `maxSnapshotsPerReport` default turns a transient burst into permanent staleness

```ts
// before
const snapshots = this.buffer.splice(0, this.buffer.length);
// after
const maxSnapshots = config.runtimeMetricsMaxSnapshotsPerReport ?? 1;
const snapshots = this.buffer.splice(0, Math.min(this.buffer.length, maxSnapshots));
```

`splice(0, 1)` takes from the **front** — the *oldest* snapshot — and keeps the remainder. `collectSample()` pushes one per second and the report timer now drains one per second, so a backlog never shrinks. After a 5-minute gRPC outage the buffer holds 300 samples and the agent spends the rest of the process lifetime reporting values that are 5 minutes old, one per second, never catching up. Once the buffer saturates at 600 it locks in at ~10 minutes stale.

These are gauges — heap size, RSS, CPU%. A five-minute-old heap reading isn't just useless, it's actively wrong on a dashboard because it arrives stamped as current.

**To be clear, I don't think this should simply be reverted.** The pre-PR behaviour was also broken for the backlog case, just less visibly. OAP's `MeterProcessor` collapses an entire `collect` stream into one input set with a single timestamp (last non-zero wins), and `SampleFamily.createMeterSamples` groups samples by label set — these meters carry no labels, so draining 300 snapshots into one stream produced a single `MeterEntity` holding 300 duplicate samples at one timestamp. It only looked fine because the backlog cleared in one shot.

So the premise here is right: **one snapshot per stream is the protocol-correct unit.** Just the direction is wrong.

```ts
const snapshot = this.buffer.pop();   // newest, not oldest
this.buffer.length = 0;               // stale gauges have no value — discard
if (!snapshot) { resolve(); return; }
```

Follow-on worth considering: once it's latest-wins, the 600-entry ring buffer has no job left. Both `runtimeMetricsBufferSize` and the new `runtimeMetricsMaxSnapshotsPerReport` become vestigial — a single "latest snapshot" slot covers it. I'd like this PR to end with *fewer* knobs than it started with, not one more.

## Partially revert the metadata loop

With one snapshot per stream, the rework in `MeterSender.ts` has nothing left to do. Please restore the `metadataWritten` guard — `Meter.proto` specifies these fields as "be set value in the first element in the stream-call", and setting them on every element roughly doubles the per-report payload for no gain.

Worth knowing why the current form can't work as intended anyway: `MeterProcessor.read()` does `if (data.getTimestamp() > 0) { timestamp = data.getTimestamp(); }` — last one wins, and `process()` then builds one `SampleFamily` per meter name at that single timestamp. Per-snapshot timestamps in a multi-snapshot stream silently collapse. If anyone ever raises `maxSnapshotsPerReport`, data points are lost with no error.

**Do keep `snapshot.collectedAt` as the timestamp source** — stamping with actual sample time rather than `Date.now()` at report time is a genuine accuracy improvement. That reduces this hunk to roughly a one-line change.

## Reporting `0` when heap-space detail is disabled

When `heapSpaceDetail` is false, `RuntimeSampler` sets `oldSpaceUsed`/`newSpaceUsed` to `0` and the collector still emits both meters. On a dashboard a flat zero line is indistinguishable from a genuine reading. If a meter can't be sampled, omit it from `toMeterData()` — MQE `exists` and the UI's `visibleWhen` handle absence correctly, but neither can detect a fabricated zero.

That said, I'd question the flag itself. `v8.getHeapSpaceStatistics()` returns ~7 entries and is called at 1 Hz; that isn't a cost that justifies a config option, a code branch, and a test that mutates the config singleton to reach it.

## New env vars use the deprecated prefix

`SW_AGENT_RUNTIME_METRICS_HEAP_SPACE_DETAIL` and `SW_AGENT_RUNTIME_METRICS_MAX_SNAPSHOTS_PER_REPORT` both land on `SW_AGENT_RUNTIME_METRICS_*`, which the README explicitly documents as a deprecated alias family. Canonical is `SW_AGENT_NODEJS_RUNTIME_METRICS_*` — every existing entry uses it. Whichever options survive should move.

Also, the PR description mentions "env documentation", but neither variable was added to the README env table — only the meter table changed.

## `instance_nodejs_new_space_used: gt 0` will flake

In `tests/plugins/express/expected.data.yaml`. V8's new space `space_used_size` is legitimately `0` immediately after a scavenge, so this will fail intermittently in CI. Use `ge 0`, as you correctly did for `array_buffers`. Same argument, more weakly, for `malloced_memory`.

## Backend and UI follow-ups

Agent-only is a fine scope for this PR, but the new meters won't reach users until two follow-ups land, and it'd be good to track them as linked issues so the wire format and MAL rules don't drift:

- `apache/skywalking` — `meter-analyzer-config/nodejs-runtime.yaml` declares exactly the original six `metricsRules`. Without new entries MAL never converts the new meters, so nothing is stored or queryable.
- `apache/skywalking-horizon-ui` — `general.json` has panels for the same six.

## Smaller things

- `RuntimeSampler` now imports the global config singleton and re-reads the flag on every sample. Read it once in the constructor or inject it — that's also what forces the test to `require()` and mutate a module singleton to exercise the branch. Moot if the flag goes away.
- `config.runtimeMetricsHeapSpaceDetail !== false` — the parser always yields a boolean, so `!== false` is redundant.
- `mockCollect` in `tests/remote/MeterSender.test.ts` is declared and cleared in `beforeEach` but never used.
- `it('clears reporter stub on DISCONNECT')` never asserts the stub was cleared — it only checks the client constructor was called once. Assert `reporterClient` is `undefined`, or rename the test.
- `toMeterData: jest.fn(() => [mockMeterData, { ...mockMeterData }])` — the spread copies the same `jest.fn` references, so it's one mock object twice. Counts still work, but it reads as two distinct meters.
- The CI job is named `TestUnitRemote` but runs `tests/remote/ tests/config/ tests/runtime/`. `TestUnit` would be more accurate, and it could use a `name:` field to match the `Lint` job.

## One question

Twelve meters at 1 Hz per instance doubles meter write volume on OAP. `instance_nodejs_uptime` in particular is a deterministic ramp — useful for spotting restarts, but not at one-second resolution. Do you think the report period should vary per meter, or is that better left until the OAP-side rules land and we can see the actual load?

Nice work overall — the `process.cpuUsage()` change is a good catch. The old code called `cpuUsage()` twice with a gap between the two calls, losing whatever CPU time elapsed in between.

songzhendong:
Thanks for the professional and detailed review.

Addressed the MeterSender / config / E2E points as follows.

### Fix

- **Latest-wins**: keep a single `latestSnapshot` slot (overwrite on collect, clear after report); drop the ring buffer.
- **Fewer knobs**: remove `runtimeMetricsBufferSize` / `runtimeMetricsMaxSnapshotsPerReport` and related env / deprecated aliases.
- **Metadata**: restore `metadataWritten` — service / instance / timestamp only on the first stream element; keep `snapshot.collectedAt` as the timestamp.
- **Heap-space flag**: remove `runtimeMetricsHeapSpaceDetail`; always sample heap spaces (no fabricated zeros via a disabled flag).
- **Env / README**: surviving knobs use the canonical `SW_AGENT_NODEJS_RUNTIME_METRICS_*` prefix and stay in the README table; deleted knobs are gone rather than renamed.
- **E2E**: in `tests/plugins/express/expected.data.yaml`, `instance_nodejs_new_space_used` and `instance_nodejs_malloced_memory` use `ge 0` (same rationale as `array_buffers`).

### On report period / write volume

You asked whether report period should vary per meter, or wait until OAP-side rules land and we can see the actual load.

My preference for this PR is **safety first: lower the default report frequency now**, rather than keep 12×1 Hz until we have measured OAP load.

Concretely:

- Default **collect / report period is now 20s** for the shared runtime meters — aligned with Java’s `meter.report_interval` (Meter path), not the JVM 1s package.
- **`instance_nodejs_uptime` defaults to every 30s** (`SW_AGENT_NODEJS_RUNTIME_METRICS_UPTIME_REPORT_PERIOD`), since it is a deterministic ramp and does not need second-level resolution.
- Users can still raise frequency via env / `agent.start()` if they need finer dashboards.



### OAP / UI follow-ups

Companion work for the six new meters is already implemented on the fork (MAL `nodejs-runtime.yaml` + Horizon General instance panels). I will open / update the apache PRs and **pin the Node agent commit** once this agent change is ready to land, so the wire format and UI stay in sync.

### Test / Verification

- Unit: `MeterSender`, `AgentConfig`, `RuntimeMetricsCollector`, `RuntimeSampler`
- Express expected assertions updated for `ge 0`; plugin tests override periods to 1s so CI stays fast
- Local source stack verified: all **12** `meter_instance_nodejs_*` series present (including the six new ones)

tests/plugins/express/client.ts
```diff
@@ -33,12 +37,12 @@ app.use('/test', testRouter);
...
-    r.on('data', (chunk) => (data += chunk));
-    r.on('end', () => res.send(data));
-  })
-  .end();
+    .request(`http://${process.env.SERVER || 'localhost:5000'}${req.url}`, (r) => {
+      let data = '';
+      r.on('data', (chunk) => (data += chunk));
+      r.on('end', () => res.send(data));
+    })
+    .end();
```
wu-sheng: nit: this reindent is formatting-only and unrelated to the metrics change — could you revert it? It adds 6 lines of noise to an otherwise focused diff.

`server.ts` got the same `agent.start()` addition without the reindent, and the repo's prettier scripts are scoped to `src/` only (`"format": "prettier --write \"src/**/*.ts\""`, same for `lint:fix` and `lint`), so `tests/` has never been prettier-formatted — this looks like editor format-on-save rather than an intentional pass. If tests should be covered by prettier, that's worth its own PR with the `package.json` scripts change behind it.

.github/workflows/test.yaml
```diff
@@ -47,6 +47,34 @@ jobs:
...
+
+      - name: Set Up NodeJS ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Unit tests on Node@${{ matrix.node-version }}
+        run: |
+          npm i
+          npx jest tests/remote/ tests/config/ tests/runtime/ --runInBand
```
wu-sheng: Good addition — worth calling out that this is the first thing in CI to actually run `tests/config/`, `tests/remote/`, and `tests/runtime/`. Those suites have existed on master and never executed, since `TestPlugins` only runs `tests/plugins/<plugin>/`.

Two suggestions:

**1. Name.** `TestUnitRemote` describes one of the three directories, but this is the repo's *complete* unit-test run — those three are exactly all the non-plugin `*.test.ts` files. `TestUnit` would be accurate:

```yaml
  TestUnit:
```

No `name:` field needed — `build-matrix`, `TestPlugins`, and `TestLib` all omit it, so `TestUnit (20)` reads consistently next to `TestPlugins (20, express)`.

**2. Select by exclusion rather than an allowlist.**

```yaml
          npx jest --testPathIgnorePatterns "/tests/plugins/" --runInBand
```

The hardcoded directory list preserves the exact failure mode this job fixes: add `tests/foo/` later and it silently never runs, the same way `tests/config/` and `tests/remote/` sat dead until now. Plugins are the only suites needing docker, so excluding them is the stable rule. Moving `testPathIgnorePatterns` into `jest.config.js` would also make a bare local `jest` behave the same way, but the CLI flag is the one-line version.

tests/plugins/express/client.ts
songzhendong: Reverted the unrelated reindent.

.github/workflows/test.yaml
songzhendong: Renamed to TestUnit and switched to --testPathIgnorePatterns "/tests/plugins/".

wu-sheng (commented):
Follow-up commit looks right — `TestUnit` with `--testPathIgnorePatterns`, and the reindent reverted. 37/37 checks green.

One correction to my own CI suggestion: `--testPathIgnorePatterns` *replaces* jest's default `["/node_modules/"]` rather than appending, so node_modules is no longer excluded from test discovery. It's green today because nothing in the current dependency tree matches jest's test patterns, but a future dep could change that. Safer form:

```yaml
npx jest --testPathIgnorePatterns "/node_modules/" "/tests/plugins/" --runInBand
```

## One last thing: collapse to a single timer and drop the uptime knob

These are the same change, so I'll take them together.

**Two periods stopped meaning anything when the ring buffer went away.** Under the old design, collect 1s + report 10s genuinely meant "batch ten samples per RPC" — the buffer carried them. With a single `latestSnapshot` slot, collecting faster than reporting just overwrites and discards, and reporting faster than collecting sends nothing at all. The second period is dead weight this PR created.

**There's a real defect hiding in it.** `RuntimeSampler` computes CPU% as a delta between consecutive `sample()` calls, so the measurement window is the *collect* period. Set collect 1s / report 20s and each published `instance_nodejs_process_cpu` describes only the final second before the report — the other 19 seconds of CPU history are discarded, and a noisy spot reading is presented as the interval's value. One timer makes the CPU window exactly equal to the reporting interval, with no gaps or overlap.

```ts
private timer?: NodeJS.Timeout;

boot(): void {
  if (this.timer) { logger.warn('MeterSender timer already scheduled; skipping duplicate boot.'); return; }
  this.timer = setInterval(() => {
    if (this.closed) return;
    this.collectSample();
    void this.reportBufferedMetrics();
  }, config.runtimeMetricsReportPeriod || 20000) as NodeJS.Timeout;
  this.timer.unref();
}
```

**And `runtimeMetricsUptimeReportPeriod` goes with it.** The 1s → 20s default already answered the volume question — 20x across every meter, uniformly. The per-meter override is a second answer to a question the first one closed, and it wasn't re-examined after 20s landed in the same commit. It buys about 4% of meter volume and costs `lastUptimeReportAt`, `shouldIncludeUptime()`, `DEFAULT_UPTIME_REPORT_PERIOD_MS`, the `gauges.splice(7, 0, ...)`, an untested env var, and a `config` import in what should be a pure mapper — the same coupling we removed from `RuntimeSampler` in round one.

Removing it is a revert, not a redesign: `2a1feaf7` already had uptime inline between `array_buffers` and `peak_malloced_memory`. Restore that line and the splice, the gate, and the state field all disappear.

Net effect is deletion — two config options and their `nvm*` aliases out of `AgentConfig` and the normalize/apply/clear helpers, one README row, and the overrides in `AgentConfig.test.ts`, `MeterSender.test.ts`, `client.ts`, `server.ts`. Nothing to deprecate: `git show v0.9.0:src/config/AgentConfig.ts` has no `nvm` fields, so none of this has ever shipped.

## Otherwise

Everything else from both rounds is done — latest-wins reporting, the metadata guard per the proto contract, dead knobs removed, canonical env prefix, `ge 0` on the flaky assertions, and a CI job that finally runs unit suites which had been sitting dead on master since they were written.

Dismissing my earlier request for changes; it's stale. Nothing above blocks merge if you'd rather land this and do the timer consolidation separately — but it's mostly deletion and it's all in code this PR already touches, so folding it in seems cleaner than a follow-up.

songzhendong:
Thanks for the follow-up.

Updated TestUnit to
npx jest --testPathIgnorePatterns "/node_modules/" "/tests/plugins/" --runInBand
so the CLI ignore list does not drop Jest’s default node_modules exclusion.

Collapsed collect/report into a single timer on runtimeMetricsReportPeriod (default 20s): each tick runs collectSample() then reportBufferedMetrics(), so the CPU delta window matches the publish interval.

Removed runtimeMetricsCollectPeriod, runtimeMetricsUptimeReportPeriod, and the related env / nvm collect aliases. instance_nodejs_uptime is included with the other gauges on every report again.

wu-sheng (approved):

wu-sheng:
Please continue on OAP and UI side as follow-up.

PR #144, opened 2026-08-10
feat(grpc): native multi-backend failover via grpc-js

songzhendong:
## Summary

Native grpc-js multi-backend failover for `SW_AGENT_COLLECTOR_BACKEND_SERVICES`.

### Behavior

- **One** `host:port` → plain target (grpc-js `dns:` — multi-IP, re-resolve, natural TLS authority)
- **Multiple** (comma-separated) → `sw-static:///` + `pick_first`, reconnect backoff (1s→30s), `UNAVAILABLE` retry
- No app-level poll/reselect; no gRPC keepalive (stock OAP rejects frequent idle pings)
- `grpc.enable_http_proxy: 0` (host HTTP proxy must not break multi-address / is unsupported for agent→OAP)
- Auth rejection: throttled error log only (no backend rotation)
- READY→IDLE treated as disconnect; coalesced flush for Trace/Meter

### Tests / CI

- Unit: `tests/remote/`
- E2E: `tests/remote-e2e/static-failover/` (deterministic active/standby failover)
- CI: `TestUnit` ignores `remote-e2e`; dedicated `TestRemoteE2E` job

### Out of scope

- Java-style periodic multi-name DNS expand
- Agent→OAP via HTTP proxy
- CommandService / TLS multi-SAN automation beyond README guidance

### Test plan

- [x] Unit: `npx jest tests/remote/ --runInBand`
- [x] E2E: `tests/remote-e2e/static-failover/`
- [x] Fork CI green on https://github.com/songzhendong/skywalking-nodejs/pull/14 @ `4d6e292`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -73,7 +73,7 @@ jobs:
       - name: Unit tests on Node@${{ matrix.node-version }}
         run: |
           npm i
-          npx jest --testPathIgnorePatterns "/node_modules/" "/tests/plugins/" --runInBand
+          npx jest --testPathIgnorePatterns "/node_modules/" "/tests/plugins/" "/tests/remote-e2e/" --runInBand
 
   build-matrix:
     runs-on: ubuntu-latest
@@ -118,6 +118,37 @@ jobs:
         run: |
           npm i
           npm run test tests/plugins/${{ matrix.plugin }}/
+
+
+  TestRemoteE2E:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    strategy:
+      matrix:
+        node-version: [ 20, 22, 24 ]
+    env:
+      SW_NODE_VERSION: ${{ matrix.node-version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: Set Up NodeJS ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Remote E2E On Node@${{ matrix.node-version }} (static-failover)
+        run: |
+          npm i
+          npm run test tests/remote-e2e/static-failover/
 
   TestLib:
     runs-on: ubuntu-latest
--- a/docs/en/advanced/troubleshooting.md
+++ b/docs/en/advanced/troubleshooting.md
@@ -15,7 +15,8 @@ logs.
 6. Check that the path, suffix, and HTTP method are not filtered by trace settings.
 7. Check the agent log for plugin version or OAP connection errors.
 
-If the process is short-lived, call `await agent.flush()` before exit.
+If the process is short-lived, call `await agent.flush()` before exit. `flush()` is a bounded
+best-effort attempt, not a delivery guarantee. See [Data is missing when the process stops](#data-is-missing-when-the-process-stops).
 
 ## OAP connection errors
 
@@ -25,8 +26,12 @@ The collector setting must use `host:port` form:
 export SW_AGENT_COLLECTOR_BACKEND_SERVICES=oap.example.com:11800
 ```
 
-Check DNS, network access, firewall rules, and the OAP gRPC port. The current agent uses only the
-first item if the value is a comma-separated list.
+Check DNS, network access, firewall rules, and the OAP gRPC port.
+
+One `host:port` uses the grpc-js `dns:` resolver (all A/AAAA records become endpoints and are
+re-resolved periodically). A comma-separated list uses a static resolver with `pick_first`: each
+name is a literal endpoint only — no DNS expansion or re-resolution per name, so discovery is weaker
+than a single DNS name (for example a headless Kubernetes service).
 
 An option passed to `agent.start()` replaces the environment value. This includes an empty string:
 
@@ -39,11 +44,26 @@ For a TLS OAP endpoint, set `SW_AGENT_SECURE=true`. The agent uses the system tr
 configuration for a custom CA or client certificate. Set `SW_AGENT_AUTHENTICATION` if OAP requires
 an agent token.
 
+Under TLS with multiple hostnames, certificate verification follows the channel authority (the first
+list entry in the configured target). Endpoint pick order may be shuffled by grpc-js, but the target
+string — and therefore authority / SNI — stays in config order. Every backend must present a
+certificate that shares the needed SANs, or failover handshakes fail. Prefer one DNS name with
+multiple A/AAAA records for TLS high availability.
+
+Channel disconnect lines are logged at `error` and recover lines at `warn` (throttled separately so a
+recover line is not swallowed by the disconnect window). For per-address grpc-js detail, set
+`GRPC_TRACE=pick_first,subchannel`.
+
 ## Traces are missing during an OAP outage
 
 The agent keeps finished segments in a memory buffer. When the buffer reaches
-`SW_AGENT_MAX_BUFFER_SIZE`, it removes the oldest finished segment. Restore the OAP connection;
-increasing the buffer only delays data loss and uses more process memory.
+`SW_AGENT_MAX_BUFFER_SIZE`, it removes the oldest finished segment and may log that the trace buffer
+reached maximum size. Restore the OAP connection; increasing the buffer only delays data loss and
+uses more process memory.
+
+If a report attempt fails, the agent may log that it discarded N trace segment(s) after report
+failure. Those segments are not re-sent. Reporting is best-effort: failures discard data, and
+`flush()` only waits briefly then tries once more.
 
 ## A library has no spans
 
@@ -65,11 +85,27 @@ See [Node.js runtime metrics](../features/runtime-metrics.md) for meter names an
 
 ## Agent logs are hard to find
 
-The default agent log level is `error`.
+The default agent log level is `warn`.
+
+| Level | What you typically see |
+| --- | --- |
+| `error` | Auth rejection, channel disconnect (throttled), fatal boot failures |
+| `warn` (default) | Channel recovered (throttled), trace buffer full, discarded segment batches, meter report failures (throttled) |
+| `info` / `debug` | Lifecycle noise; span debug lines |
+
+Levels below the configured threshold are silent by design (`warn` / `info` / `debug` become no-ops
+when the threshold is higher).
 
 - When `NODE_ENV` is not `production`, logs go to the console.
 - When `NODE_ENV=production`, logs go to `skywalking.log` in the process working directory.
 - Set `SW_LOGGING_TARGET=console` to use the console in production.
+
+## Agent to OAP over an HTTP proxy
+
+Agent to OAP over an HTTP proxy is not supported. Every gRPC channel sets `grpc.enable_http_proxy=0`,
+so host `http_proxy` / `https_proxy` never affect OAP uplink. This applies to single-address and
+multi-address targets. After upgrade, traffic that previously relied on an HTTP CONNECT proxy to
+reach OAP will no longer use that proxy.
 
 ## Data is missing when the process stops
 
@@ -79,6 +115,9 @@ The default agent log level is `error`.
 await agent.flush();
 agent.destroy();
 ```
+
+`flush()` waits a short time for in-flight work and may start one more report attempt. It does not
+guarantee delivery if OAP is slow or unreachable.
 
 Do not use stop and restart as a normal agent update method inside one process. Module patches stay
 installed after `destroy()`.
--- a/docs/en/setup/configuration.md
+++ b/docs/en/setup/configuration.md
@@ -24,7 +24,7 @@ Node.js process. `SW_DISABLE` is checked when `start()` is called.
 | --- | --- | --- | --- |
 | `SW_AGENT_NAME` | `serviceName` | `your-nodejs-service` | Service name shown in SkyWalking. |
 | `SW_AGENT_INSTANCE` | `serviceInstance` | Host name | Service instance name shown in SkyWalking. |
-| `SW_AGENT_COLLECTOR_BACKEND_SERVICES` | `collectorAddress` | `127.0.0.1:11800` | OAP gRPC address in `host:port` form. |
+| `SW_AGENT_COLLECTOR_BACKEND_SERVICES` | `collectorAddress` | `127.0.0.1:11800` | OAP gRPC address(es). One `host:port` uses grpc-js `dns:` (all A/AAAA become endpoints; periodically re-resolved). A comma-separated list uses a static resolver with `pick_first` (literal endpoints only; no per-name DNS expansion or re-resolution). Endpoint pick order is shuffled by grpc-js (`shuffleAddressList`); the target string keeps config order so TLS authority / SNI stay on the first list entry. Under TLS, all backends must present certificates that share the needed SANs. Prefer one DNS name with multiple A/AAAA records for TLS high availability. |
 | `SW_AGENT_SECURE` | `secure` | `false` | Use TLS for the OAP gRPC connection. |
 | `SW_AGENT_AUTHENTICATION` | `authorization` | Not set | Authentication token sent to OAP. |
 | `SW_AGENT_TRACE_TIMEOUT` | `traceTimeout` | `10000` | gRPC deadline in milliseconds for trace and meter reports and service management requests. Must be a positive integer. |
@@ -35,14 +35,12 @@ For token authentication, set the same token in OAP with `SW_AUTHENTICATION`. Se
 When `secure` is enabled, the agent uses the system trust store. It does not provide options for a
 custom CA, client certificate, or mutual TLS.
 
-Use one OAP address. If a comma-separated list is set, the current agent uses only the first entry.
-
 ## Agent control and logging
 
 | Environment variable | `agent.start()` option | Default | Description |
 | --- | --- | --- | --- |
 | `SW_DISABLE` | None | Not set | Set the exact value `true` to keep the agent stopped. |
-| `SW_AGENT_LOGGING_LEVEL` | None | `error` | Agent log level: `error`, `warn`, `info`, or `debug`. |
+| `SW_AGENT_LOGGING_LEVEL` | None | `warn` | Agent log level: `error`, `warn`, `info`, or `debug`. |
 | `SW_LOGGING_TARGET` | None | See below | Set to `console` to log to the console in production. |
 | `SW_AGENT_MAX_BUFFER_SIZE` | `maxBufferSize` | `1000` | Limit for active and buffered trace segments. Must be a positive integer. |
 | `SW_AGENT_DISABLE_PLUGINS` | `disablePlugins` | Empty | Comma-separated plugin file names without the `Plugin` suffix, such as `mysql,express`. |
--- a/src/agent/core/boot/ServiceManager.ts
+++ b/src/agent/core/boot/ServiceManager.ts
@@ -51,7 +51,7 @@ class ServiceManager {
       try {
         service.prepare();
       } catch (error) {
-        logger.error('ServiceManager prepare failed: %s', error);
+        logger.error(`ServiceManager prepare failed: ${error}`);
       }
     }
 
@@ -59,7 +59,7 @@ class ServiceManager {
       try {
         service.boot();
       } catch (error) {
-        logger.error('ServiceManager boot failed: %s', error);
+        logger.error(`ServiceManager boot failed: ${error}`);
       }
     }
 
@@ -67,7 +67,7 @@ class ServiceManager {
       try {
         service.onComplete();
       } catch (error) {
-        logger.error('ServiceManager onComplete failed: %s', error);
+        logger.error(`ServiceManager onComplete failed: ${error}`);
       }
     }
 
@@ -79,7 +79,7 @@ class ServiceManager {
       try {
         service.shutdown();
       } catch (error) {
-        logger.error('ServiceManager shutdown failed: %s', error);
+        logger.error(`ServiceManager shutdown failed: ${error}`);
       }
     }
     this.bootedServices.clear();
--- a/src/agent/core/meter/MeterSender.ts
+++ b/src/agent/core/meter/MeterSender.ts
@@ -28,9 +28,16 @@ import { RuntimeSnapshot } from './RuntimeSampler';
 import GRPCChannelManager from '../remote/GRPCChannelManager';
 import { GRPCChannelListener } from '../remote/GRPCChannelListener';
 import { GRPCChannelStatus } from '../remote/GRPCChannelStatus';
+import {
+  coalesceReport,
+  flushCoalesced,
+  FLUSH_WAIT_MS,
+  ReportCoalesceState,
+  runCollectStream,
+} from '../remote/coalesceReport';
 
 const logger = createLogger(__filename);
-const logReportError = throttled(logger, 'error', 30000);
+const logReportError = throttled(logger, 'warn', 30000);
 
 /** Reports Node.js runtime metrics via gRPC MeterReportService (Go/Python-compatible pipeline). */
 export default class MeterSender implements BootService, GRPCChannelListener {
@@ -41,7 +48,7 @@ export default class MeterSender implements BootService, GRPCChannelListener {
   /** Latest gauge snapshot only — stale samples have no value after reconnect. */
   private latestSnapshot?: RuntimeSnapshot;
   private timer?: NodeJS.Timeout;
-  private reporting?: Promise<void>;
+  private readonly reportState: ReportCoalesceState = {};
 
   private collector!: RuntimeMetricsCollector;
 
@@ -75,12 +82,7 @@ export default class MeterSender implements BootService, GRPCChannelListener {
     if (!this.channelManager) {
       return undefined;
     }
-
-    return new MeterReportServiceClient(
-      config.collectorAddress,
-      grpc.credentials.createInsecure(),
-      this.channelManager.getClientOptions(),
-    );
+    return this.channelManager.createClient(MeterReportServiceClient);
   }
 
   private startTimer(): void {
@@ -95,83 +97,61 @@ export default class MeterSender implements BootService, GRPCChannelListener {
   }
 
   private collectSample(): void {
+    // Always sample so RuntimeSampler CPU deltas stay on the report period (not the outage length).
     this.latestSnapshot = this.collector.sample();
   }
 
   private reportBufferedMetrics(): Promise<void> {
+    return coalesceReport(
+      this.reportState,
+      () => this.doReportBufferedMetrics(),
+      () => this.closed,
+    );
+  }
+
+  private doReportBufferedMetrics(): Promise<void> {
     if (this.closed) {
       return Promise.resolve();
     }
 
-    if (this.reporting) {
-      return this.reporting;
+    if (!config.serviceName || !config.serviceInstance) {
+      return Promise.resolve();
     }
 
-    this.reporting = this.doReportBufferedMetrics().finally(() => {
-      this.reporting = undefined;
-    });
+    // Check connectivity before consuming the snapshot so disconnect-window samples are kept.
+    if (this.status !== GRPCChannelStatus.CONNECTED || !this.reporterClient) {
+      return Promise.resolve();
+    }
 
-    return this.reporting;
-  }
+    const snapshot = this.latestSnapshot;
+    if (!snapshot) {
+      return Promise.resolve();
+    }
+    this.latestSnapshot = undefined;
 
-  private doReportBufferedMetrics(): Promise<void> {
-    return new Promise((resolve) => {
-      try {
-        if (this.closed) {
-          resolve();
-          return;
-        }
-
-        const snapshot = this.latestSnapshot;
-        this.latestSnapshot = undefined;
-        if (!snapshot || this.status !== GRPCChannelStatus.CONNECTED || !this.reporterClient) {
-          resolve();
-          return;
-        }
-
-        if (!config.serviceName || !config.serviceInstance) {
-          resolve();
-          return;
-        }
-
-        const stream = this.reporterClient.collect(
-          new grpc.Metadata(),
-          { deadline: Date.now() + (config.traceTimeout || 10000) },
-          (error: grpc.ServiceError | null) => {
-            if (error) {
-              logReportError('Failed to report runtime meter data', error);
-              this.reportGrpcError(error);
-            }
-            resolve();
-          },
-        );
-
-        try {
-          let metadataWritten = false;
-          for (const meterData of this.collector.toMeterData(snapshot)) {
-            // Meter.proto: service / instance / timestamp on the first stream element only.
-            if (!metadataWritten) {
-              meterData
-                .setService(config.serviceName)
-                .setServiceinstance(config.serviceInstance)
-                .setTimestamp(snapshot.collectedAt);
-              metadataWritten = true;
-            }
-            stream.write(meterData);
+    const client = this.reporterClient;
+    const serviceName = config.serviceName;
+    const serviceInstance = config.serviceInstance;
+    return runCollectStream({
+      open: (onStatus) =>
+        client.collect(new grpc.Metadata(), { deadline: Date.now() + (config.traceTimeout || 10000) }, onStatus),
+      writeAll: (stream) => {
+        let metadataWritten = false;
+        for (const meterData of this.collector.toMeterData(snapshot)) {
+          // Meter.proto: service / instance / timestamp on the first stream element only.
+          if (!metadataWritten) {
+            meterData.setService(serviceName).setServiceinstance(serviceInstance).setTimestamp(snapshot.collectedAt);
+            metadataWritten = true;
           }
-        } finally {
-          try {
-            stream.end();
-          } catch (error) {
-            logReportError('Failed to end meter collect stream', error);
-            resolve();
-          }
+          stream.write(meterData);
         }
-      } catch (error) {
-        logReportError('Failed to report runtime meter data', error);
+      },
+      onFailure: (reason, error) => {
+        logReportError(reason, error);
         this.reportGrpcError(error);
-        resolve();
-      }
+      },
+      openFailureReason: 'Failed to report runtime meter data',
+      endFailureReason: 'Failed to end meter collect stream',
     });
   }
 
@@ -183,12 +163,21 @@ export default class MeterSender implements BootService, GRPCChannelListener {
     this.channelManager?.reportError(error);
   }
 
+  /**
+   * Best-effort: one shared FLUSH_WAIT_MS budget for forceReport of latest snapshot, then in-flight wait.
+   */
   flush(): Promise<any> | null {
     if (this.closed) {
       return null;
     }
     this.collectSample();
-    return this.reportBufferedMetrics();
+    return flushCoalesced(
+      this.reportState,
+      () => this.doReportBufferedMetrics(),
+      () => this.closed,
+      () => this.latestSnapshot != null,
+      FLUSH_WAIT_MS,
+    );
   }
 
   shutdown(): void {
@@ -197,7 +186,7 @@ export default class MeterSender implements BootService, GRPCChannelListener {
       clearInterval(this.timer);
       this.timer = undefined;
     }
-    this.reporting = undefined;
+    this.reportState.reporting = undefined;
     this.reporterClient = undefined;
     this.latestSnapshot = undefined;
     this.collector.destroy();
--- a/src/agent/core/remote/BackendAddressResolver.ts
+++ b/src/agent/core/remote/BackendAddressResolver.ts
@@ -1,0 +1,161 @@
+/*!
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import * as grpc from '@grpc/grpc-js';
+import { createLogger } from '../../../logging';
+
+const logger = createLogger(__filename);
+
+const SW_STATIC_SCHEME = 'sw-static';
+
+let swStaticResolverRegistered = false;
+
+/**
+ * Parse one host:port via grpc.experimental.splitHostPort.
+ * Returns normalized host:port or null when invalid (logged at error).
+ */
+export function tryParseHostPort(entry: string): string | null {
+  const trimmed = entry.trim();
+  if (!trimmed) {
+    return null;
+  }
+  const parsed = grpc.experimental.splitHostPort(trimmed);
+  if (!parsed?.host || parsed.port == null) {
+    logger.error(`Invalid collector address: ${entry}`);
+    return null;
+  }
+  if (parsed.port <= 0 || parsed.port > 65535) {
+    logger.error(`Invalid collector address (bad port): ${entry}`);
+    return null;
+  }
+  const host = parsed.host.includes(':') ? `[${parsed.host}]` : parsed.host;
+  return `${host}:${parsed.port}`;
+}
+
+/** Parse comma-separated backend host:port entries. Invalid entries are logged and dropped. */
+export function parseStaticBackendAddresses(raw: string): string[] {
+  const result: string[] = [];
+  for (const part of raw.split(',')) {
+    const normalized = tryParseHostPort(part);
+    if (normalized) {
+      result.push(normalized);
+    }
+  }
+  return result;
+}
+
+/**
+ * Build a grpc-js channel target.
+ * - One address: plain host:port (default dns: resolver — multi-IP + periodic re-resolve + natural TLS authority).
+ * - Multiple: sw-static:/// list for pick_first across explicit backends (literal endpoints only;
+ *   no DNS expansion / re-resolution of each name — weaker discovery than a single dns: target).
+ */
+export function buildNativeGrpcTarget(addresses: string[]): string {
+  // Caller (openChannel) already requires a non-empty list.
+  if (addresses.length === 1) {
+    return addresses[0]!;
+  }
+  ensureSwStaticResolverRegistered();
+  return `${SW_STATIC_SCHEME}:///${addresses.join(',')}`;
+}
+
+function ensureSwStaticResolverRegistered(): void {
+  if (swStaticResolverRegistered) {
+    return;
+  }
+  const { registerResolver, statusOrFromValue, statusOrFromError, splitHostPort: grpcSplit } = grpc.experimental;
+
+  /**
+   * Static multi-address resolver for comma-separated backends.
+   * Mirrors grpc-js resolver-ip.js (static endpoint list, no DNS expansion).
+   */
+  class SwStaticResolver {
+    private readonly listener: grpc.experimental.ResolverListener;
+    private readonly endpoints: grpc.experimental.Endpoint[];
+    private readonly error: { code: number; details: string; metadata: grpc.Metadata } | null;
+    private hasReturnedResult = false;
+
+    constructor(
+      target: grpc.experimental.GrpcUri,
+      listener: grpc.experimental.ResolverListener,
+      _channelOptions: grpc.ChannelOptions,
+    ) {
+      this.listener = listener;
+      this.endpoints = [];
+      this.error = null;
+
+      if (target.scheme !== SW_STATIC_SCHEME) {
+        this.error = {
+          code: grpc.status.UNAVAILABLE,
+          details: `Unrecognized scheme ${target.scheme} in sw-static resolver`,
+          metadata: new grpc.Metadata(),
+        };
+        return;
+      }
+
+      const pathList = target.path
+        .split(',')
+        .map((p) => p.trim())
+        .filter(Boolean);
+      const endpoints: grpc.experimental.Endpoint[] = [];
+      for (const path of pathList) {
+        const hp = grpcSplit(path);
+        if (!hp?.host || hp.port == null) {
+          this.error = {
+            code: grpc.status.UNAVAILABLE,
+            details: `Failed to parse sw-static address ${path}`,
+            metadata: new grpc.Metadata(),
+          };
+          return;
+        }
+        endpoints.push({ addresses: [{ host: hp.host, port: hp.port }] });
+      }
+      this.endpoints = endpoints;
+    }
+
+    updateResolution(): void {
+      if (this.hasReturnedResult) {
+        return;
+      }
+      this.hasReturnedResult = true;
+      process.nextTick(() => {
+        if (this.error) {
+          this.listener(statusOrFromError(this.error), {}, null, '');
+        } else {
+          this.listener(statusOrFromValue(this.endpoints), {}, null, '');
+        }
+      });
+    }
+
+    destroy(): void {
+      this.hasReturnedResult = false;
+    }
+
+    static getDefaultAuthority(target: grpc.experimental.GrpcUri): string {
+      const first = target.path.split(',')[0]?.trim();
+      if (!first) {
+        throw new Error('sw-static target path must contain at least one host:port');
+      }
+      return first;
+    }
+  }
+
+  registerResolver(SW_STATIC_SCHEME, SwStaticResolver);
+  swStaticResolverRegistered = true;
+}
--- a/src/agent/core/remote/GRPCChannel.ts
+++ b/src/agent/core/remote/GRPCChannel.ts
@@ -18,7 +18,7 @@
  */
 
 import * as grpc from '@grpc/grpc-js';
-import { ClientOptions, connectivityState } from '@grpc/grpc-js';
+import { ClientOptions, connectivityState, ChannelOptions } from '@grpc/grpc-js';
 import ChannelBuilder, { ChannelBuildContext } from './ChannelBuilder';
 import ChannelDecorator from './ChannelDecorator';
 
@@ -26,31 +26,41 @@ export default class GRPCChannel {
   private readonly originChannel: grpc.Channel;
   private readonly interceptors: grpc.Interceptor[];
 
-  private constructor(host: string, port: number, channelBuilders: ChannelBuilder[], decorators: ChannelDecorator[]) {
+  /**
+   * Builders must spread `context.options` when they replace the options object
+   * so native channel options (keepalive / service_config) are preserved.
+   * `extraOptions` are merged exactly once into the initial context.
+   */
+  private constructor(
+    target: string,
+    channelBuilders: ChannelBuilder[],
+    decorators: ChannelDecorator[],
+    extraOptions: ChannelOptions,
+  ) {
     let context: ChannelBuildContext = {
       credentials: grpc.credentials.createInsecure(),
-      options: {},
+      options: { ...extraOptions },
     };
 
     for (const builder of channelBuilders) {
       context = builder.build(context);
     }
 
-    this.originChannel = new grpc.Channel(`${host}:${port}`, context.credentials, context.options);
+    this.originChannel = new grpc.Channel(target, context.credentials, context.options);
     this.interceptors = decorators.map((decorator) => decorator.build());
   }
 
   static create(
-    host: string,
-    port: number,
+    target: string,
     channelBuilders: ChannelBuilder[],
     decorators: ChannelDecorator[],
+    extraOptions: ChannelOptions = {},
   ): GRPCChannel {
-    return new GRPCChannel(host, port, channelBuilders, decorators);
+    return new GRPCChannel(target, channelBuilders, decorators, extraOptions);
   }
 
-  static newBuilder(host: string, port: number): GRPCChannelBuilder {
-    return new GRPCChannelBuilder(host, port);
+  static newBuilder(target: string): GRPCChannelBuilder {
+    return new GRPCChannelBuilder(target);
   }
 
   getChannel(): grpc.Channel {
@@ -68,20 +78,23 @@ export default class GRPCChannel {
     return this.originChannel.getConnectivityState(requestConnection) === connectivityState.READY;
   }
 
+  getConnectivityState(requestConnection = false): connectivityState {
+    return this.originChannel.getConnectivityState(requestConnection);
+  }
+
   shutdownNow(): void {
     this.originChannel.close();
   }
 }
 
 class GRPCChannelBuilder {
-  private readonly host: string;
-  private readonly port: number;
+  private readonly target: string;
   private readonly channelBuilders: ChannelBuilder[] = [];
   private readonly decorators: ChannelDecorator[] = [];
+  private extraOptions: ChannelOptions = {};
 
-  constructor(host: string, port: number) {
-    this.host = host;
-    this.port = port;
+  constructor(target: string) {
+    this.target = target;
   }
 
   addManagedChannelBuilder(builder: ChannelBuilder): this {
@@ -94,7 +107,12 @@ class GRPCChannelBuilder {
     return this;
   }
 
+  withChannelOptions(options: ChannelOptions): this {
+    this.extraOptions = { ...this.extraOptions, ...options };
+    return this;
+  }
+
   build(): GRPCChannel {
-    return GRPCChannel.create(this.host, this.port, this.channelBuilders, this.decorators);
+    return GRPCChannel.create(this.target, this.channelBuilders, this.decorators, this.extraOptions);
   }
 }
--- a/src/agent/core/remote/GRPCChannelManager.ts
+++ b/src/agent/core/remote/GRPCChannelManager.ts
@@ -18,11 +18,12 @@
  */
 
 import * as grpc from '@grpc/grpc-js';
-import { ClientOptions } from '@grpc/grpc-js';
+import { ClientOptions, ChannelOptions, ChannelCredentials } from '@grpc/grpc-js';
 import config from '../../../config/AgentConfig';
-import { createLogger } from '../../../logging';
+import { createLogger, throttled } from '../../../logging';
 import AgentIDDecorator from './AgentIDDecorator';
 import AuthenticationDecorator from './AuthenticationDecorator';
+import { buildNativeGrpcTarget, parseStaticBackendAddresses } from './BackendAddressResolver';
 import GRPCChannel from './GRPCChannel';
 import { GRPCChannelListener } from './GRPCChannelListener';
 import { GRPCChannelStatus } from './GRPCChannelStatus';
@@ -31,57 +32,95 @@ import StandardChannelBuilder from './StandardChannelBuilder';
 import TLSChannelBuilder from './TLSChannelBuilder';
 
 const logger = createLogger(__filename);
+const logAuthRejected = throttled(logger, 'error', 30000);
+const logChannelDisconnected = throttled(logger, 'error', 30000);
+const logChannelRecovered = throttled(logger, 'warn', 30000);
+
+/** Placeholder authority — ignored when channelOverride is set (grpc-js client.js). */
+const STUB_AUTHORITY = 'skywalking-backend';
+
+function isGrpcAuthError(error: unknown): boolean {
+  const code = (error as grpc.ServiceError | undefined)?.code;
+  return code === grpc.status.PERMISSION_DENIED || code === grpc.status.UNAUTHENTICATED;
+}
 
 function isGrpcNetworkError(error: unknown): boolean {
   const code = (error as grpc.ServiceError | undefined)?.code;
-
+  if (isGrpcAuthError(error)) {
+    return false;
+  }
   return (
     code === grpc.status.UNAVAILABLE ||
-    code === grpc.status.PERMISSION_DENIED ||
-    code === grpc.status.UNAUTHENTICATED ||
     code === grpc.status.RESOURCE_EXHAUSTED ||
-    code === grpc.status.UNKNOWN
+    code === grpc.status.UNKNOWN ||
+    code === grpc.status.DEADLINE_EXCEEDED
   );
 }
 
+function nativeChannelOptions(): ChannelOptions {
+  // No gRPC keepalive: stock OAP (grpc-java) rejects frequent idle pings with GOAWAY
+  // ENHANCE_YOUR_CALM. Agent traffic (trace/heartbeat/metrics) provides liveness.
+  //
+  // Agent→OAP via HTTP proxy is not supported. Disable grpc-js proxy interception
+  // uniformly (including single-address dns: targets that grpc-js could otherwise
+  // proxy) so host-app http_proxy/https_proxy cannot affect the agent channel —
+  // multi-address targets are also broken under HTTP CONNECT (proxy sees the
+  // unresolved comma list).
+  return {
+    'grpc.enable_http_proxy': 0,
+    'grpc.initial_reconnect_backoff_ms': 1_000,
+    'grpc.max_reconnect_backoff_ms': 30_000,
+    'grpc.service_config': JSON.stringify({
+      // shuffleAddressList: distribute agents across backends without rewriting the
+      // target string (keeps channel authority / SNI stable under TLS).
+      loadBalancingConfig: [{ pick_first: { shuffleAddressList: true } }],
+      // Retry only unary idempotent reportInstanceProperties. Client-streaming
+      // collect (trace/meter) must not retry — grpc-js replays the write buffer and
+      // OAP does not dedupe segments. keepAlive is covered by the next 20s tick.
+      // FQDN matches Management.proto (skywalking.v3), not ManagementCompat.
+      methodConfig: [
+        {
+          name: [{ service: 'skywalking.v3.ManagementService', method: 'reportInstanceProperties' }],
+          retryPolicy: {
+            maxAttempts: 3,
+            initialBackoff: '1s',
+            maxBackoff: '10s',
+            backoffMultiplier: 2,
+            retryableStatusCodes: ['UNAVAILABLE'],
+          },
+        },
+      ],
+    }),
+  };
+}
+
 /**
- * Shared gRPC channel manager (Java GRPCChannelManager skeleton).
- * V1: single address; V2 reserved: multi-address failover via reportError().
+ * Shared gRPC channel manager using grpc-js native multi-address failover
+ * (pick_first + reconnect backoff + service-config retry). Opens one channel at boot.
  */
 export default class GRPCChannelManager implements BootService {
   private managedChannel: GRPCChannel | null = null;
   private readonly listeners: GRPCChannelListener[] = [];
   private lastStatus: GRPCChannelStatus | null = null;
+  private lastConnectivityState: grpc.connectivityState | null = null;
   private closed = false;
-
-  /** V1: first address when comma-separated; V2: failover selection. */
-  resolveAddress(): string {
-    const raw = config.collectorAddress ?? '';
-    const first = raw.split(',')[0]?.trim();
-    if (!first) {
-      throw new Error('collectorAddress is not configured');
-    }
-    return first;
-  }
-
-  getChannel(): grpc.Channel {
-    if (!this.managedChannel) {
-      throw new Error('gRPC channel is not available');
-    }
-
-    return this.managedChannel.getChannel();
-  }
+  private grpcServers: string[] = [];
 
   getClientOptions(): ClientOptions {
     if (!this.managedChannel) {
       throw new Error('gRPC channel is not available');
     }
-
     return this.managedChannel.getClientOptions();
   }
 
-  isConnected(): boolean {
-    return this.managedChannel?.isConnected(true) ?? false;
+  /**
+   * Construct a generated gRPC client bound to the shared channel.
+   * Address/credentials are placeholders — transport uses channelOverride.
+   */
+  createClient<TClient>(
+    ClientCtor: new (address: string, credentials: ChannelCredentials, options?: ClientOptions) => TClient,
+  ): TClient {
+    return new ClientCtor(STUB_AUTHORITY, grpc.credentials.createInsecure(), this.getClientOptions());
   }
 
   addChannelListener(listener: GRPCChannelListener): void {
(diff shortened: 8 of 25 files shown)


wu-sheng:
Thanks for this PR — multi-backend failover is a capability the Node.js agent genuinely needs, and several pieces here are solid regardless of approach (the bare-IPv6 fix, the MeterSender simplification, the E2E infrastructure).

Before this merges, I'd like to propose a direction change on the transport core. The goal here is retry/failover *capability*, and grpc-js (already at 1.14.4 in our lockfile) ships that capability natively. I verified each piece against the vendored source:

**Feature-by-feature mapping** (existing + added by this PR → native mechanism):

| Capability | This PR | Native grpc-js |
|---|---|---|
| Multi-backend list | hand-parsed list + per-backend channel rebuild | `ipv4:a:port,b:port` multi-address target — the IP resolver splits on commas (`resolver-ip.js:54`); hostname lists via a small custom resolver registered with `experimental.registerResolver` (~50 lines, modeled on `resolver-dns`) |
| Failover trigger | 30s poll + error classification + random reselect | transport-level: subchannel `TRANSIENT_FAILURE` → `pick_first` deterministically tries the next address immediately (`load-balancer-pick-first.js:282`), per-address exponential backoff built in |
| Dead/black-holed backend | not detected (channel stays READY; DEADLINE_EXCEEDED is unclassified) | `grpc.keepalive_time_ms` / `grpc.keepalive_timeout_ms` / `grpc.keepalive_permit_without_calls` (`channel-options.js:31-32`) |
| Reconnect pacing | `SW_AGENT_COLLECTOR_GRPC_CHANNEL_CHECK_INTERVAL` + `SW_AGENT_FORCE_RECONNECTION_PERIOD` | `grpc.initial_reconnect_backoff_ms` / `grpc.max_reconnect_backoff_ms` (`channel-options.js:35-36`) |
| Per-RPC retry | none — a failed batch is dropped (buffer cleared in `finally`) | service-config `retryPolicy` (`retrying-call.js`) via the `grpc.service_config` channel option, plus `waitForReady` metadata so calls queue during a failover window instead of failing fast (`resolving-call.js:168`) |
| Load spreading across agents (Java's random pick) | `Math.random` reselect | shuffle the address list once before building the target string |
| Random initial selection / sticky backend | random + sticky | `pick_first` is sticky by design; `round_robin` available if spreading is ever preferred |
| TLS | per-selected-host channel | same `createSsl()` on the one channel; set `grpc.ssl_target_name_override`/`grpc.default_authority` (`channel-options.js:26,29`) — note this is one name for all backends, fine when OAP nodes share a cert/SAN |
| Auth token / Agent-Version headers | interceptors | unchanged — interceptors are orthogonal to resolver/LB (and call-credentials can't be used on insecure channels: `channel-credentials.js:103`) |
| Status notify → stub recreation | listener cascade + generation guards | the channel is never replaced, so stubs are created once; keep a thin `watchConnectivityState` watcher only if timer gating is still wanted |

**Sketch of the native core** (replaces most of `GRPCChannelManager`):

```ts
const addresses = shuffle(parseBackendAddresses(config.collectorAddress));
const target = addresses.every(isIpLiteral)
  ? `ipv4:${addresses.join(',')}`
  : `sw:///${addresses.join(',')}`;      // custom resolver scheme for hostnames

const channel = new grpc.Channel(target, credentials, {
  'grpc.keepalive_time_ms': 30_000,
  'grpc.keepalive_timeout_ms': 10_000,
  'grpc.keepalive_permit_without_calls': 1,
  'grpc.initial_reconnect_backoff_ms': 1_000,
  'grpc.max_reconnect_backoff_ms': 30_000,
  'grpc.service_config': JSON.stringify({
    loadBalancingConfig: [{ pick_first: {} }],
    methodConfig: [{
      name: [{}],
      retryPolicy: {
        maxAttempts: 3, initialBackoff: '1s', maxBackoff: '10s',
        backoffMultiplier: 2, retryableStatusCodes: ['UNAVAILABLE'],
      },
    }],
  }),
});
```

(Streaming-call retries are bounded by grpc-js's per-RPC retry buffer, so the segment batch win comes mainly from `waitForReady` queuing plus connection-level failover being near-instant.)

**Why I'm raising this now rather than reviewing the hand-rolled core line by line**: a deep pass over this PR found six confirmed correctness issues in the polling state machine — DEADLINE_EXCEEDED/black-hole never triggers failover; auth-rejected backends pin the agent permanently; a transient per-RPC error on a healthy READY channel arms a reselect that tears the healthy channel down (and cancelled calls from the closed channel can flip it back to the dead backend); CONNECTING is treated as down, so a slow handshake gets aborted every check tick; `grpcChannelCheckInterval: 0` via programmatic config spins `setInterval` at ~1ms; and the empty-list recovery branch is unreachable because `boot()` returns before starting the timer. Four of these six have no place to exist in the native model, because application code no longer owns channel lifecycle. The two new env knobs this PR introduces haven't shipped in any release yet, so redirecting now has zero compatibility cost.

**What still needs fixing on either path**:

1. `BackendAddressResolver` silently drops invalid entries (no log at any level; master accepted `oap:11800/` via `parseInt` and otherwise surfaced `Invalid collector address: X` at error level) — parsing stays for target construction, so this stays relevant.
2. Auth-failure failover needs a small explicit handler either way (per-RPC UNAUTHENTICATED doesn't change connectivity state, so no LB policy reacts to it): on persistent auth errors, rotate the rejecting address to the back of the list and rebuild the channel (~20 lines, rare path).
3. The static-failover E2E passes ~50% of runs without exercising failover (the agent's initial backend is random; the test stops collector-a and only asserts collector-b received data). Make it deterministic: find the active collector via `/receiveData`, stop *that one*, assert the *other* got the post-stop ping.
4. `MeterSender`: `flush()` returns the stale in-flight promise without sending the just-collected snapshot (in Lambda that invocation's metrics are lost); and the snapshot is consumed before the CONNECTED check, discarding disconnect-window samples — check status before consuming, and chain a drain after the in-flight report.
5. E2E helpers: `docker ps -aq --filter name=testcontainers | xargs docker rm -f` force-kills every testcontainers-managed container on the shared daemon, and Ryuk is disabled via a module-import side effect — testcontainers 10.3.2's `compose.getContainer(...).stop()` does the failover kill cleanly with Ryuk left on.
6. CI: the TestRemoteE2E job pre-builds the agent image and then `ensureE2eAgentImage()` force-rebuilds the same tag inside jest (`forceRebuild` when CI is set) — every matrix cell builds twice; also `--build-arg NPM_REGISTRY` is dead (no `ARG` in Dockerfile.agent), and `needs: [ build-matrix ]` serializes the job for outputs it never reads.

Happy to help review the native-core revision; the mapping table above is meant to make that switch mostly mechanical.

songzhendong:
Thanks for the detailed review — agreed on the direction.

I've reworked this PR to use **grpc-js native multi-address** instead of the hand‑rolled poll/reselect channel manager:

- Multi‑address target via `ipv4:` / `ipv6:` / custom `sw‑static` resolver
- `pick_first` load balancing
- Keepalive + reconnect backoff
- `service_config` retry (`UNAVAILABLE`) + `waitForReady`
- Dropped the unused check‑interval / force‑reconnection knobs (not released yet)

The previously noted hand‑rolled issues should no longer apply under this model (blackhole/`DEADLINE_EXCEEDED` tear‑down, READY transient rebuild, treating `CONNECTING` as down, `checkInterval: 0`, unreachable empty‑list recovery).

Also addressed the must‑fix items on either path:

1. **Address parse logging** — invalid entries are logged at error level (not silently dropped)
2. **Auth rejection** — `UNAUTHENTICATED` / `PERMISSION_DENIED` rotates the rejecting backend to the end and rebuilds the channel (native LB does not failover on auth status codes)
3. **Deterministic remote E2E** — resolve active collector → stop it → assert traffic on standby
4. **MeterSender flush** — only consume after CONNECTED; pending drain after in‑flight report
5. **E2E helpers** — no global `docker rm -f` / Ryuk disable; stop via compose service container
6. **CI** — avoid double agent image build; drop dead `NPM_REGISTRY`; `TestRemoteE2E` no longer depends on `build‑matrix`

wu-sheng:
Thanks for the quick turnaround — the native rework checks out: I verified the multi-address target, `pick_first`, keepalive, backoff and retry service config against the grpc-js source and they are all wired correctly; the E2E is now deterministic, the helpers use proper testcontainers APIs, invalid addresses are logged, and the unreleased knobs are gone.

One design decision first, then a full list of changes. **Everything below should be addressed — no optional items.**

**Design: remove the auth-rotation machinery entirely.** Backends in one cluster share the same token, so switching backends cannot fix an auth rejection. If a token refresh is rolling through the cluster, waiting briefly is the correct behavior — the channel stays healthy and the next report succeeds once the OAP accepts. Unreachable-backend failover needs no extra code either: `pick_first` already tries the next address deterministically, and `grpc.initial_reconnect_backoff_ms`/`max_reconnect_backoff_ms` are the reconnect cooldown. Removing rotation also removes a crash in the current revision: a mid-run rebuild swaps channels with no status transition (CONNECTED→CONNECTED is deduped in `notify()`), so all three clients keep stubs bound to the shut-down channel, and the heartbeat's unguarded stub call turns grpc-js's synchronous `Error('Channel has been shut down')` into an uncaught exception that **crashes the host application** within one heartbeat period.

Changes:

1. On `UNAUTHENTICATED`/`PERMISSION_DENIED`: keep the channel, emit a **throttled** error-level log pointing at `SW_AGENT_AUTHENTICATION`, and do nothing else. Delete `rotateBackendAndRebuild`, `currentAuthority`, `moveAddressToEnd`, the `rebuilding` flag, `rebuildForTest()`, and their tests.
2. Wrap `sendHeartbeat()`'s stub calls (`keepAlive`, `reportInstanceProperties`) in try/catch routed to `reportGrpcError`, like the other two clients — grpc-js Client methods can throw synchronously, and an APM agent must never crash its host.
3. Logging: connectivity transitions currently produce no log at all, and the default `SW_AGENT_LOGGING_LEVEL` is `error`, so the existing `warn`/`debug` lines are invisible by default. Add, at a level visible under defaults: one line on DISCONNECT (with the configured backend list, noting reconnection with backoff), one line on recovery, plus the throttled auth line. Add a README note that per-address connection detail is available via grpc-js's `GRPC_TRACE=pick_first,subchannel`.
4. `flush()` completion guarantee, fixed once in a shared helper for all clients: `TraceSegmentServiceClient.reportOnce()` returns the in-flight promise with no follow-up drain, and `MeterSender`'s `pendingDrain` re-drain is detached (`void`), so a flush during an in-flight report resolves before the new data is sent. Chain instead: `if (this.reporting) return this.reporting.then(() => this.report...());`.
5. Delete `watcherGeneration` — every `openChannel()` installs a new `GRPCChannel` object, so the existing `this.managedChannel !== managed` identity check already covers every invalidation.
6. Replace the decorative stub address with a manager-owned factory: grpc-js ignores the address/credentials arguments whenever `channelOverride` is set (`client.js:57`), so `resolveAddress()` feeds a dead parameter. Add `GRPCChannelManager.createClient(Ctor)` passing a constant placeholder authority + `getClientOptions()`, and delete `resolveAddress()` and the triplicated stub-construction blocks.
7. `GRPCChannel`: merge `extraOptions` exactly once (the post-builder re-merge is a no-op today and has ambiguous precedence if a builder ever replaces the options object); make the builder contract explicit.
8. `buildNativeGrpcTarget`: always emit `sw-static:///` — the custom resolver handles IP and hostname lists identically, so the `ipv4:`/`ipv6:` special cases (and the untested ipv6/mixed branches, and the `isIpv4Literal`/`isIpv6Literal` wrappers) can go.
9. Use `grpc.experimental.splitHostPort`/`combineHostPort` instead of the hand-rolled parsers, and collapse validate/normalize/build into a single `tryParseHostPort()` pass — each address is currently parsed three times, and two grammars for the same string can drift.
10. `openChannel()`: pass `false` to `notifyCurrentConnectivityState` — `watchConnectivityState()` already called `getConnectivityState(true)`, so the second connection request is redundant.
11. E2E: replace the ~33s of fixed `sleep()`s with `waitForExpect` polling on the conditions they wait for (the file already uses the pattern); this saves ~99s of CI wall-clock per PR across the 3-version matrix and removes flake risk.
12. Keep exactly one copy of the agent-image build: drop the workflow's "Pre-build E2E agent image" step (the helper already builds on `docker image inspect` miss), and slim the Dockerfile to `COPY package*.json` + one `npm ci` — the compose file mounts host `src/` over `/app/src`, so the image's second `npm install` and protoc output are built and then masked.
13. `docker-compose.yml`: define collector-a/b via `extends` from `base-compose.yml`'s collector service (move the host-port mapping out of the base) — the mock-collector SHA is currently pinned in three places.
14. `upCompose()`: drop the two collector `Wait.forHealthCheck` strategies — the server's `depends_on: condition: service_healthy` already gates on both collectors, so waiting on the server alone suffices.

With #1/#5/#6/#8 applied, `GRPCChannelManager` shrinks to: parse + shuffle once, open one channel at boot, watch connectivity for logging/stub gating, throttled auth log, shutdown. That is the right end state for the native model.

songzhendong:
Thanks for the Round-2 review.



### Fix

1. **Auth**: on `UNAUTHENTICATED`/`PERMISSION_DENIED`, keep the channel and emit a throttled error log pointing at `SW_AGENT_AUTHENTICATION`. Removed `rotateBackendAndRebuild`, `currentAuthority`, `moveAddressToEnd`, `rebuilding`, and related tests.
2. **Heartbeat**: `keepAlive` / `reportInstanceProperties` wrapped in try/catch → `reportGrpcError`.
3. **Logging + README**: error-level DISCONNECT (backends + reconnect/backoff) and recovery logs; README notes `GRPC_TRACE=pick_first,subchannel` for per-address detail.
4. **Flush**: chain drains via `reporting.then(() => report…())` in Trace / Meter clients (same completion guarantee; not extracted to a shared helper).
5. **Removed** `watcherGeneration` — identity check on `managedChannel` is enough.
6. **`createClient(Ctor)`** on the manager; deleted `resolveAddress()` and duplicated stub construction.
7. **`GRPCChannel`**: merge `extraOptions` once with explicit builder precedence.
8. **Always** `sw-static:///`; dropped `ipv4:`/`ipv6:` branches.
9. **Parse** via `grpc.experimental.splitHostPort`; local `combineHostPort` only because it is not re-exported on `experimental`.
10. **`notify(..., false)`** after `watchConnectivityState()`.
11. **E2E**: fixed sleeps replaced with `waitForExpect` polling.
12. **CI/Dockerfile**: dropped workflow Pre-build; single `npm ci`. Kept `COPY .` (not package-only) because plugin E2E does not mount `src/` into the image — remote-e2e still bind-mounts host `src/`.
13. **Compose**: collectors `extends` base; host port moved out of base into plugin compose files.
14. **`upCompose()`**: wait on server health only.

### Verification

Fork CI (Build / License / Test) was green **40/40** on  
https://github.com/songzhendong/skywalking-nodejs/pull/14/checks @ `fb1e237`.

This PR is updated to the same squashed commit `fb1e237`.

wu-sheng:
Round-2 items all check out — verified the rotation removal, heartbeat guards, logging, and cleanups against the tree; unit tests pass and compilation is clean. This round went one layer deeper — into the native channel options block and its interaction with real environments (grpc-js internals, a stock OAP's server policy, proxy env vars; several items were reproduced live). A few of these correct suggestions from my earlier reviews, noted inline. As before, everything below should be addressed — no optional items.

**Design decisions (with the reasoning to keep in code comments / README):**

1. **Disable grpc-js proxy interception**: add `'grpc.enable_http_proxy': 0` to the channel options, with a comment and a README note explaining why. grpc-js decides proxying once per channel on the *unresolved* target string (HTTP CONNECT model — the proxy does the resolving), so it runs before our resolver, cannot split the comma list, and emits a malformed `CONNECT a:11800,b:11800:443` — with `https_proxy` set in the host app's env (very common), the agent silently loses all connectivity (reproduced live). Per-address proxying does not exist in grpc-js, and agent→OAP via HTTP proxy has never been a supported scenario — this flag makes the agent explicitly immune to host-app proxy env vars instead of accidentally broken by them.
2. **Delete all three keepalive options** (`keepalive_time_ms`, `keepalive_timeout_ms`, `keepalive_permit_without_calls`). A stock OAP configures no `permitKeepAlive*` on its gRPC server, so grpc-java's defaults apply (one idle ping per 2h, two strikes) — the current 30s idle pings draw `GOAWAY ENHANCE_YOUR_CALM (too_many_pings)` ~90s after connect, killing in-flight collect streams in a repeating churn cycle (and the mock-collector in this PR's e2e is also grpc-java, so the test can flake on it). The pings are unnecessary: the HTTP/2 connection persists on its own, and the agent generates steady traffic (1s trace reports, 20s heartbeat, 20s runtime metrics) that serves as liveness. The Java agent configures no gRPC keepalive either — its only "keepAlive" is the application-level `InstancePingPkg` heartbeat. (The 30s values came from my round-1 sketch; that suggestion was wrong.)
3. **Remove `waitForReady: true` from the service config** — grpc-js validates and stores `methodConfig.waitForReady` but never reads it at runtime; only per-call metadata options are honored, and all calls here use plain `new grpc.Metadata()`. It is dead configuration that misleads readers about failover semantics.
4. **TLS + multi-backend, zero new code**: when the parsed list has exactly **one** entry, bypass `sw-static` and use the plain `host:port` target (grpc-js default `dns:` resolver) — a single hostname resolving to multiple IPs then gets full native behavior (all IPs resolved, tried in order, re-resolved on reconnect) with a correct TLS authority, which is the recommended HA shape under TLS. Document in README: for TLS HA, prefer one DNS name → multiple IPs; comma-listing *distinct* hostnames under TLS requires certificates sharing SANs, because grpc-js derives SNI/verification for every connection from the channel authority (the first list entry).
5. **`DEADLINE_EXCEEDED` on a READY channel stays unhandled — intentionally.** Failover targets unreachable backends, not functional health of a connected one. Please add a one-line comment in `reportError` stating this is deliberate, so a future maintainer doesn't "fix" it.

**Behavior fixes:**

6. **IDLE swallow loses a trace batch at outage onset**: `notifyCurrentConnectivityState` skips IDLE, but READY→IDLE is grpc-js's normal path after the active connection drops — listeners keep status CONNECTED, the next 1s tick writes the whole buffer into a doomed call and clears it in `finally`. Master preserved the buffer via DISCONNECT; MeterSender got the check-before-consume guard, the trace client didn't. Fix: notify DISCONNECT on IDLE-after-READY (keep the CONNECTING exemption), and/or don't clear the trace buffer until the collect callback confirms delivery.
7. **Flush coalescing**: `return this.reporting.then(() => this.reportOnce())` chains one follow-up per concurrent caller — K flushes during one in-flight report queue K serialized cycles. Replace with a single shared follow-up (`this.followUp ??= this.reporting.then(...)`, cleared when it runs), extracted into one helper used by both clients. (The per-caller chaining is the literal shape of my round-2 suggestion; it needed the coalesced form.)
8. **False error on clean shutdown**: `shutdown()` routes through `notify(DISCONNECT)`, logging error-level "disconnected …; reconnecting with exponential backoff" on every deliberate `agent.destroy()`. Guard the log (not the listener fan-out) with `this.closed`; while there, the boot-time first transition can also log "disconnected … reconnecting" for an agent that never connected — move the log wording into a small status-transition helper.
9. **Make the failover E2E self-healing**: after stopping the active collector, the test sends exactly one ping+flush then polls the standby for 120s with no new traffic — if that single segment dies in the failover window (finding 6 makes that likely on a loaded runner), the test can never pass. Move `pingServer()` + `flushServer()` inside the `waitForExpect` retry loop.
10. **Stale local image cache**: `ensureE2eAgentImage()` skips the build whenever the tag exists, keyed only on node version — after a dependency bump, the e2e runs fresh bind-mounted `src/` against old baked `node_modules`. Key the tag on a content hash (package-lock.json + Dockerfile), or switch to compose `build:` and let docker layer caching handle it.
11. **Rewrite the PR description** — it still documents the deleted poll/reselect design, including two env knobs (`SW_AGENT_COLLECTOR_GRPC_CHANNEL_CHECK_INTERVAL`, `SW_AGENT_FORCE_RECONNECTION_PERIOD`) that no longer exist; operators reading the merged PR would configure no-ops.

**Cleanup:**

12. Delete the dead exports: `splitHostPort` (tagged `@deprecated` in the same PR that adds it; only consumer is its own test), `formatHostPort`, and the unused `sleep` helper; drop the unreachable `?? 443` default in `combineHostPort` (or deep-import the original from `@grpc/grpc-js/build/src/uri-parser`, which is exported there); add a one-line provenance comment on `SwStaticResolver` ("mirrors grpc-js resolver-ip.js") so grpc-js upgrades trigger a re-diff.
13. Remove the `as any` in `registerResolver(SW_STATIC_SCHEME, SwStaticResolver as any)` — the class satisfies `ResolverConstructor` today, and the cast only converts a future compile-time break into runtime "channel never resolves"; make `getDefaultAuthority` throw on an empty path per the documented contract instead of returning `''`.
14. Restore `ports: 12800:12800` to the base collector in `base-compose.yml` and give static-failover a port-less variant (second base service or inlined 7-line collector defs) — the current diff copy-pastes the ports stanza into 9 plugin compose files, and the 10th plugin author who misses it gets a 300s-timeout mystery failure.
15. Delete `getGrpcServersForTest()` — the tests already capture the target string passed to `GRPCChannel.newBuilder` and can assert shuffle membership/no-rebuild on it.
16. `stopComposeService`: testcontainers 10.x drives compose v2 naming, so the three-candidate loop with error-message sniffing reduces to ``await compose.getContainer(`${service}-1`).stop()``.
17. The "does not treat CONNECTING as DISCONNECT" test passes vacuously if `boot()` no-ops — add a positive assertion (`newBuilder`/`watchConnectivityState` called).
18. The single-entry matrix `case: [ static-failover ]`: hardcode the path until a second case exists, or derive it from the directory like `build-matrix` does for plugins.

With 1–3 applied, the hardcoded options shrink to reconnect backoff (1s→30s) and `retryPolicy` (3× UNAVAILABLE) — sensible defaults that need no new config knobs. No further design concerns from my side; after this pass the transport core is where it should be.

songzhendong:
Thanks for the Round-3 review.



### Fix

**Design**
1. **`grpc.enable_http_proxy: 0`** on the channel, with code comment + README note (host `http(s)_proxy` must not break multi-address targets; agent→OAP via HTTP proxy is unsupported).
2. **Removed all gRPC keepalive options** — stock OAP rejects frequent idle pings with `GOAWAY ENHANCE_YOUR_CALM`; agent traffic provides liveness.
3. **Removed `waitForReady` from service config** — dead config in grpc-js unless set per-call.
4. **Single address → plain `host:port` (`dns:`)**; multiple → `sw-static:///`. README documents TLS HA (prefer one DNS → many IPs; distinct hostnames need shared SANs).
5. **`DEADLINE_EXCEEDED` on READY** left unhandled on purpose — comment in `reportError` so it is not “fixed” later.

**Behavior**
6. **READY→IDLE → DISCONNECT** (CONNECTING still exempt). Trace batch is restored on failed collect instead of clearing in `finally`.
7. **Flush coalescing** via shared `coalesceReport()` (single follow-up for Trace + Meter).
8. **Shutdown logging**: connectivity error logs skipped when `closed`; wording distinguishes never-connected vs reconnect.
9. **Failover E2E**: `ping` + `flush` + assert inside `waitForExpect` after stopping the active collector.
10. **Agent image**: remote-e2e uses compose `build:` so Docker layer cache invalidates on lockfile/Dockerfile change (dropped tag-only `ensureE2eAgentImage`).
11. **PR description** updated to the native design (no poll/reselect env knobs).

**Cleanup**
12. Removed dead exports (`splitHostPort` / `formatHostPort` / unused `sleep`); dropped `?? 443` on `combineHostPort`; provenance comment on `SwStaticResolver` (mirrors `resolver-ip.js`). Local `combineHostPort` kept because it is not re-exported on `experimental`.
13. Removed `as any` on `registerResolver`; `getDefaultAuthority` throws on empty path.
14. Restored `ports: 12800` on base collector; added `collector-no-host-port` for remote-e2e (no more ports copy-paste across plugin compose files).
15. Removed `getGrpcServersForTest()`; tests assert the `newBuilder` target string.
16. `stopComposeService` only uses `` `${service}-1` ``.
17. CONNECTING test asserts `newBuilder` / `watchConnectivityState` were called.
18. CI hardcodes `tests/remote-e2e/static-failover/` (no single-element `case` matrix).


### Verification

Fork CI (Build / License / Test) is green **40/40** on  
https://github.com/songzhendong/skywalking-nodejs/pull/14/checks @ `4d6e292`.

This PR is updated to the same squashed commit `4d6e292`.

wu-sheng:
All 18 round-3 items landed — verified the substantive ones directly: the keepalive options are gone, a single-address list now uses the plain `dns:` target, and READY→IDLE correctly notifies DISCONNECT (guarded on the previous state being READY). Branch compiles and lints clean, `tests/remote/` passes 22/22.

This round's findings sit almost entirely in the error paths of those fixes, and several trace back to my own round-3 guidance — noted inline where that's the case. Two design decisions first; everything below should be addressed.

**Design decisions:**

1. **Trace buffer: discard on send failure; never re-send.** Remove all three `this.buffer.unshift(...batch)` restore sites in `doReport`. Round 3 asked for keep-on-failure, but that was the wrong half of the fix: the data that actually needs protecting is the disconnect-window data, and the IDLE notification fix already protects it — a known-down channel is caught by the `status !== CONNECTED` check at line 152, so the buffer is never spliced and stays bounded by `maxBufferSize`. A failure on a *healthy* channel means the batch is gone, accepted. Restoring it instead produced four separate problems, all of which disappear with the restore:
   - a resend storm: on a slow-but-READY backend the status deliberately stays CONNECTED (per our DEADLINE_EXCEEDED decision), so the same batch was re-sent every second, amplified 3× by the retryPolicy;
   - duplicate segments: the synchronous catch paths restored and resolved while leaving the call open, so the callback later restored the *same* batch again (lines 167 / 186 / 192) — OAP received every written segment twice, and the buffer grew by a batch per failed attempt;
   - `maxBufferSize` bypass: the cap is enforced only in the 'segment-finished' listener, so restores pushed the buffer to ~2× its configured bound;
   - post-shutdown resurrection: a callback landing after `shutdown()` refilled the buffer that shutdown had just cleared.

   Please add a throttled log on the failure path reporting the discarded count (`batch.length`) — with restore gone, dropped data must at least be visible, alongside the existing buffer-full warning. While there, that warning interpolates `config.collectorAddress`, which is now the whole comma-separated list; word it for multi-backend.

2. **HTTP proxy is unsupported — keep it disabled uniformly.** `grpc.enable_http_proxy: 0` currently applies to every channel including the single-address `dns:` target that grpc-js proxies correctly, so a single-backend deployment reaching OAP through an `http_proxy` today will go offline after this release. That is the intended policy, but the code doesn't say so: the comment justifies the flag only by the multi-address parsing problem, which doesn't cover that case. Please broaden the comment to state the policy (agent→OAP via HTTP proxy is not supported; proxy interception is disabled uniformly so host-app proxy env vars cannot affect the agent), and note the behavior change in the README.

**Fixes:**

3. **`coalesceReport` never clears `followUp` on rejection** (coalesceReport.ts:37). `state.followUp` is reset only inside the onFulfilled callback, so one rejected report leaves it a permanently-rejected promise: every later flush-during-report returns that stale rejection instead of queueing a real drain, and `void this.reportOnce().finally(...)` turns it into an unhandled rejection — which terminates the host application by default on Node ≥15. Reachable because `doReport`'s `emitter.emit('segments-sent')` sits outside its try block, so any throwing listener rejects the promise. Clear `followUp` on both settlement paths (`.finally`) and catch/route the rejection. (The coalescing shape is what I asked for in round 3; the error path was never specified.)
4. **The logger has no `format.splat()`, so every `%s` prints literally** (src/logging/index.ts:33). winston is configured with `format.json()` and a `prettyPrint()` console transport; the interpolation argument is dropped into the SPLAT symbol and discarded. The operator diagnostics rounds 2 and 3 asked for therefore convey none of their information — `gRPC channel disconnected from backends [%s]`, `Failed to build gRPC channel for target [%s]: %s`, `Invalid collector address: %s` (visible by running `npx jest tests/remote/`). The `%s` style predates this PR, but this PR adds 12 new sites that depend on it. Add `winston.format.splat()` — one line that also fixes the pre-existing call sites — or switch the new sites to template literals, which is why the buffer-full warning above does print correctly.
5. **TLS authority is a random entry** (BackendAddressResolver.ts:167). `boot()` shuffles the list and `getDefaultAuthority` returns its first element, so under TLS the SNI name and validated certificate hostname differ per process restart — the agent connects on some restarts and fails the handshake on others. The README text added here says the authority is the first entry without mentioning the shuffle, so "order your list" doesn't work as a mitigation. Skip the shuffle when `config.secure` is set, and reword the README to state the actual requirement: all backends must present certificates sharing SANs. (My round-3 wording asked for the note but didn't account for the shuffle.)
6. **MeterSender's `stream.end()` failure never reaches the manager** (MeterSender.ts:159) — it logs and resolves where TraceSegmentServiceClient calls `reportGrpcError` (line 185-188). When the meter reporter is the first or only component to observe a dead session, the manager is never told and DISCONNECT waits on the connectivity watcher alone. Mirror the trace client.
7. **Heartbeat catch misattributes** (ServiceManagementClient.ts:141): every 10th tick calls `reportInstanceProperties`, but a synchronous throw there logs 'Failed to send heartbeat', and the shared 30s throttle can then suppress genuine keepAlive failures. Distinguish the two RPCs, as the inner callbacks already do.
8. **Dockerfile layer ordering** (tests/plugins/common/Dockerfile.agent:21): `npm ci` still sits behind `COPY . .`, so any source edit busts the install layer and every plugin e2e job reinstalls from scratch — and the helper comment added in this PR describes layer caching this file doesn't provide. `npm ci` is also strict about lockfile drift where `npm install` was not, so a future dependency bump that forgets to regenerate the lockfile breaks all ~15 plugin jobs at once. Use `COPY package.json package-lock.json ./` → `RUN npm ci` → `COPY . .`.
9. **`boot()` failure paths never notify** (GRPCChannelManager.ts:179): both the unparseable-address and build-throw paths log and return without calling `notify()`, so `lastStatus` stays null and listeners get no status at all. Impact is limited since listeners default to DISCONNECT, but make it explicit — `notify(DISCONNECT)` on those paths.
10. **Dead code** (GRPCChannelManager.ts:228): `openChannel` is fully synchronous, so the post-build `this.closed` re-check and its `built.shutdownNow()` branch are unreachable; likewise `buildNativeGrpcTarget`'s empty-list throw, since `openChannel` already returns on `addresses.length === 0`. Remove both so the state machine doesn't appear to have a teardown path it lacks.
11. **E2E can hang instead of failing usefully** (helpers.ts:426): `resolveActiveCollector` throws when *both* collectors hold ping data, and since it runs inside `waitForExpect` — which can't clear collector state — the condition never converges and the test burns to timeout with an opaque error. Snapshot which collector saw the first ping, or reset collector state between probes, rather than requiring mutual exclusion.
12. **Shuffle test is too weak** (GRPCChannelManager.test.ts:598): stubbing `Math.random` to 0 exercises one permutation and the assertions are two `toContain` membership checks, so a shuffle that drops or duplicates entries would pass. Test `shuffleAddresses` directly for length and multiset equality over several seeded permutations — it now also determines the TLS authority, so it deserves real coverage.

songzhendong:
Thanks for the Round-4 review.

All 12 items are addressed.

### Fix

**Design**
1. **Trace buffer**: removed all `unshift` restore paths — discard on send failure; IDLE→DISCONNECT already protects the disconnect window. Added a throttled warn with discarded `batch.length`; buffer-full wording now refers to configured backends (multi-address list).
2. **HTTP proxy policy**: broadened the channel-option comment and README — agent→OAP via HTTP proxy is not supported; `grpc.enable_http_proxy=0` applies uniformly (including single-address), so host `http_proxy`/`https_proxy` cannot affect the agent (behavior change).

**Fixes**
3. **`coalesceReport`**: clear `followUp` in `.finally`; `.catch` so a rejected report cannot leave a permanently-rejected promise / unhandled rejection.
4. **Logger**: added `winston.format.splat()` so `%s` interpolates (including the connectivity / parse diagnostics from earlier rounds).
5. **TLS authority**: skip shuffle when `SW_AGENT_SECURE=true`; README requires backends to present certificates sharing SANs (order preserved under TLS).
6. **MeterSender**: `stream.end()` failures call `reportGrpcError` (aligned with Trace).
7. **Heartbeat**: separate throttled loggers / messages for `keepAlive` vs `reportInstanceProperties` (including the outer catch).
8. **Dockerfile**: `COPY package.json package-lock.json` → `npm ci --ignore-scripts` → `COPY .` → `npm rebuild grpc-tools && npm run prepare` (layer cache on lockfile; `--ignore-scripts` because `prepare`/`protoc` need the source tree, and only `grpc-tools` needs its install script for the protoc binary).
9. **`boot()` failure paths**: explicit `notify(DISCONNECT)` when addresses are empty or channel build fails.
10. **Dead code**: removed the unreachable post-build `closed` re-check and the empty-list throw in `buildNativeGrpcTarget`.
11. **E2E**: `resolveActiveCollector` no longer requires permanent mutual exclusion — if both collectors already have ping data, snapshot collector-a so `waitForExpect` can converge.
12. **Shuffle tests**: exercise `shuffleAddresses` directly for length + multiset equality over seeded permutations; TLS path asserts order is preserved.

### Verification

Fork CI (Build / License / Test) is green on  
https://github.com/songzhendong/skywalking-nodejs/pull/14/checks @ `bf382f0`.

wu-sheng:
Round-4 items all landed — verified each: the three `unshift` restores are gone with discard logging carrying `batch.length`, the buffer-full warning is reworded for multi-backend, the shuffle is skipped under `config.secure`, all three `boot()` failure paths notify DISCONNECT, the Dockerfile installs deps before copying sources, the unreachable `closed` re-check and empty-list throw are removed, heartbeat vs instance-properties errors log separately, MeterSender reports `stream.end()` failures to the manager, the README marks the proxy change explicitly, and the shuffle test now checks length, multiset equality and non-mutation across several seeded permutations. Thanks — that was a clean pass.

One design decision, then the remaining items.

**Design decision: `flush()` means "one attempt at everything currently buffered", not "guaranteed delivery".**

The agent is best-effort by design — a failed send discards and never re-sends, so statistics can't be double-counted (round 4). `flush()` inherits that. The problem in the current revision isn't that a send failed; it's that data is **never attempted**: `flush()` resolves while segments sit in the buffer. Two cases:

- the empty-buffer branch returns `null` without consulting `reportState.reporting` — but the buffer is empty *precisely because* an in-flight report already spliced its batch, so flush reports "nothing to do" at the moment data is in flight;
- `state.followUp` is cleared in the trailing `.finally()`, after the recursive report completes, so a flush arriving during that second report gets the existing promise and no further drain is scheduled — its data stays buffered.

Both disappear with a simpler shape. Please **delete `followUp` entirely**, reducing `coalesceReport` to a single-flight guard (if a report is in flight, return it; otherwise start one), and make `flush()` two explicit steps:

```ts
async flush(): Promise<unknown> | null {
  if (this.closed) return null;
  if (this.timeout) { clearTimeout(this.timeout); this.timeout = undefined; }
  // 1. wait for an in-flight report (its batch is already out of the buffer)
  if (this.reportState.reporting) await this.reportState.reporting.catch(() => undefined);
  // 2. send whatever is left, once — failures discard, as everywhere else
  if (this.buffer.length > 0) await this.reportOnce().catch(() => undefined);
  this.scheduleNextReport();
}
```

No loop, no retry, no new knobs. This is sufficient for the case that matters: `AWSLambdaTriggerPlugin.ts:96` calls `agent.flush()` automatically on every invocation, *after* yielding once so child spans finalize — so by the time flush runs, that invocation's segments are already buffered and the two steps cover all of them. Data arriving *during* step 2 is not sent, which is correct under best-effort: the next timer tick owns it.

Worth noting that this coalescing logic has produced a finding in each of the last three rounds (no follow-up → follow-up without error handling → follow-up cleared too late). Deleting it removes the class. Please also add a test for "flush while a report is in flight" asserting a second report ran and the buffer drained — that specific interleaving has never been covered.

**Remaining fixes:**

1. **The scheduled report's rejection is still unhandled** (coalesceReport.ts:54, TraceSegmentServiceClient.ts:126). The round-4 catch landed only inside the follow-up branch; the primary path returns `doReport().finally(...)` with rejection intact, and the timer discards it via `void this.reportOnce().finally(...)`. `doReport`'s `emitter.emit('segments-sent')` sits outside its try block, so a throwing listener rejects the promise — an unhandled rejection, which terminates the host application by default on Node ≥15 (reproducible on Node 24). With `followUp` gone, this is the only remaining path: make sure no caller can be handed a rejecting promise.
2. **`splat()` is applied twice** (logging/index.ts:33 and :42) — once at logger level and again on the Console transport. Reproduced against the vendored winston: the double config emits `{ '0':'s', '1':'w', '2':'-', …, message:'backends [a:11800,b:11800] target [sw-static:///x]' }`, i.e. the second pass spreads the already-interpolated string into per-character numeric keys on every console record. The message interpolation itself is correct — the round-4 fix worked — so just apply `splat()` once, at logger level.
3. **The E2E can still pass without exercising failover** (helpers.ts:99). Round 4's hang is fixed, but by arbitrarily labelling collector-a active when both hold ping data — which reintroduces the round-3 vacuous pass: if the channel is actually on collector-b, the test stops the idle A and then trivially finds pre-existing data on B. Make the probe unambiguous (unique per-probe operation name, or reset collector state before probing) rather than guessing; on a genuinely ambiguous state, failing loudly is better than a silent wrong answer.
4. **The synchronous catch paths double-count discards and leak the call** (TraceSegmentServiceClient.ts:190, :194). They call `discardBatch` and resolve but never `stream.end()`, so the call stays open to its deadline and its callback runs `discardBatch` for the same batch again — operators see twice the true discard count and the manager gets two error reports for one failure. Data is fine now that restore is gone; the accounting should be too. End or cancel the stream on those paths, or mark the batch already-discarded.
5. **Heartbeat attribution has an inverted edge** (ServiceManagementClient.ts:143). The outer catch branches on `reportProperties`, but when that flag is true *both* calls run inside the try — so a synchronous throw from `keepAlive` is logged as an instance-properties failure. Narrower than the original bug, but the clean fix is a try/catch around each call rather than one around both.

songzhendong:
Thanks for the Round-5 review.


### Fix

**Design**
1. Removed `followUp`. `coalesceReport` only returns the in-flight promise or starts one report.
2. `flush()` (Trace / Meter): await in-flight → report remaining buffer/snapshot once → no loop / retry. Data arriving during the second send stays for the timer.

**Fixes**
1. Report promises never reject to callers (`coalesceReport` + timer `.catch`; `segments-sent` emit wrapped so listener throws cannot become unhandled rejections on Node ≥15).
2. `splat()` applied once at logger level only (Console keeps `prettyPrint()`), so console records are not split into per-character keys.
3. E2E active-collector detection uses a unique `/probe/:token` per probe; neither → retry, both → fail loudly (no guessing collector-a).
4. Trace sync failure paths mark the batch discarded once and `cancel()` the stream so the callback cannot double-count discards / double-`reportError`.
5. Heartbeat: `reportInstanceProperties` and `keepAlive` each have their own try/catch (properties ticks still send both RPCs).

wu-sheng:
Round-5 items landed: `flush()` is the two-step shape, `discardBatch` has a once-only guard plus `stream.cancel()` with a dedicated regression test, `coalesceReport` is simplified with its own test file, and the e2e helpers were reworked. Branch type-checks clean and the unit suite passes 43/43.

This round I ran a full multi-angle sweep plus direct verification. Three of the findings below come from my own earlier suggestions — flagged as such, with corrections.

**Design decisions:**

1. **Revert `format.splat()`; use template literals instead.** My round-4 note offered `splat()` *or* template literals — `splat()` was the wrong option and should be backed out. It makes every log message a printf format string, with two consequences I reproduced against the vendored winston:

   - **User data is interpreted as a format string, and metadata is silently dropped.** Span operation names are URL paths, and any non-ASCII URL is percent-encoded. `logger.debug('Starting span /search/%d0%bf%d1%80', {traceId:'abc', spanId:1})` emits `{"message":"Starting span /search/NaN0%bf%d1%80"}` — the path is corrupted (`%d` consumed the meta object as a number) and `traceId`/`spanId` are gone entirely. This affects every `SpanContext` debug site.
   - **`%s` now expands full `Error.stack`.** Previously it printed a literal `%s` and dropped the error. This repo's own `throttled()` doc explains why that matters: it reduces errors to code/message because "a multi-KB stack … lets the records accumulate in winston's internal stream buffer faster than the transport drains them, eventually exhausting the heap." `reportError`'s two `logger.debug` calls fire on every failed RPC with **no throttle**, and `GRPCChannelListener failed: %s` is unthrottled at error level.

   Please drop `splat()` from the logger and convert the `%s` sites this PR added to template literals — which is exactly why the buffer-full warning already prints correctly today.

2. **Scope `retryPolicy` to `reportInstanceProperties` only.** It currently applies to all methods (`name: [{}]`), including the client-streaming trace/meter `collect` uploads. grpc-js's `RetryingCall` keeps a `writeBuffer` of already-sent messages and replays it on retry, with no exclusion by call type — so a batch partially consumed by OAP before an `UNAVAILABLE` can be re-sent in full, and OAP does not dedupe by segment id. That silently violates the no-duplicate-statistics rule we settled in round 4: the agent code honors it while the transport layer breaks it underneath. The retryPolicy came from my round-1 sketch and was never re-examined against that decision.

   Retry earns its place on exactly one RPC — instance properties: unary, idempotent (OAP overwrites the instance's properties), and losing it leaves instance metadata stale for a full 200s cycle. Trace/meter `collect` must be excluded (replay = duplicates); `keepAlive` needs no retry since the next tick is 20s away.

   ```ts
   methodConfig: [
     {
       name: [{ service: 'skywalking.v3.ManagementService', method: 'reportInstanceProperties' }],
       retryPolicy: {
         maxAttempts: 3,
         initialBackoff: '1s',
         maxBackoff: '10s',
         backoffMultiplier: 2,
         retryableStatusCodes: ['UNAVAILABLE'],
       },
     },
   ],
   ```

   (Please confirm whether the generated client uses `Management.proto` or `ManagementCompat.proto` so the fully-qualified service name matches.)

3. **Bound the `flush()` wait.** `flush()` now always awaits an in-flight report, which only settles at the gRPC deadline — 10s by default — when OAP is slow or black-holing. `AWSLambdaTriggerPlugin` awaits `agent.flush()` on every invocation, so an invocation with nothing new to send can stall ~10s at exit: the user pays for that duration on every call, and it can exceed the function's own timeout and fail it. An agent must never take down the app it monitors. Note `SW_AWS_LAMBDA_FLUSH` does not help here — it is a throttle ("skip if less than N seconds since the last flush"), not a timeout.

   Waiting itself is correct (the in-flight batch is already out of the buffer, so not waiting guarantees its loss), but unbounded waiting contradicts "one attempt, not guaranteed delivery" — this is the gap in the two-step shape I sketched in round 5. Race the in-flight promise against a short timer (1–2s) and return regardless when it elapses; a healthy OAP answers in milliseconds, so the bound only ever engages in the pathological case.

**Fixes:**

4. **The properties tick lost its early return** (ServiceManagementClient.ts:137). `master` had `return;` after `reportInstanceProperties` (line 130), making properties and keepAlive mutually exclusive as in the Java agent's if/else; the per-RPC try/catch restructuring dropped it, so every 10th tick now issues both RPCs. Restore the early return, or state explicitly that sending both is intended. (I saw both calls running last round but read it only as a log-attribution edge without diffing against master — my miss.)
5. **MeterSender never got the trace path's fixes** (MeterSender.ts:158). The trace client gained a `discarded` once-only flag plus `stream.cancel()`, with a regression test; MeterSender kept the old shape, so if `stream.end()` throws it logs, calls `reportGrpcError` and resolves while the live callback fires with the same error and calls `reportGrpcError` again — double DISCONNECT and throttle accounting — and the uncancelled call hangs until the deadline. Apply the same guard and cancel.
6. **`logStatusTransition` is unthrottled** (GRPCChannelManager.ts:307). Every other error site in this PR routes through `throttled()`; this one emits `logger.error` on each DISCONNECT/CONNECTED transition. A backend behind a load balancer that resets connections cycles READY→IDLE→CONNECTING→READY, producing two error lines per cycle into the winston buffer — the same accumulation regime `throttled()` exists to prevent, and on by default since the logger defaults to `error`.
7. **The disconnect-window snapshot preservation is dead code** (MeterSender.ts:189). `doReportBufferedMetrics` was reordered to keep `latestSnapshot` when disconnected, but every caller runs `collectSample()` immediately beforehand, which unconditionally overwrites it — so the preserved snapshot is never the one sent. Only a unit test poking `latestSnapshot` directly exercises the branch, which gives false confidence. Either drop the branch or have callers skip re-sampling when a preserved snapshot exists.
8. **Adding a second backend silently weakens address discovery** (BackendAddressResolver.ts:87). One address takes the grpc-js `dns:` path — all A/AAAA records become endpoints, periodically re-resolved. Two or more take `sw-static`, whose `updateResolution` returns once and yields exactly one endpoint per literal name with no re-resolution. So `oap.svc:11800` on a headless k8s service load-balances across all pod IPs and follows rescaling, while `oap-a.svc:11800,oap-b.svc:11800` freezes at two endpoints and never observes IP churn behind either name. Users adding a backend for HA get strictly weaker discovery than they had with one. Please document this in the README, and consider re-resolution in the static resolver.
9. **No boot warning for TLS with multiple hostnames** (BackendAddressResolver.ts:168). The authority pins TLS verification to the first address, so failover to a backend without that SAN fails the handshake — documented in the README but with no runtime check, so the misconfiguration surfaces only as a permanent outage during the failover it was configured for. Emit a one-time warning at boot when `config.secure && addresses.length > 1`.
10. **The e2e `/flush` deadline fallback returns 200** (tests/remote-e2e/common/server.ts:41), and the helper asserts only `status === 200` — so a hung flush is indistinguishable from a completed one. With the unbounded wait above, a regression would make every flush quietly take 8s and still pass, degrading the failover test into a timing race. Return a distinguishable status/body for the deadline path and assert on it.
11. **One discard is reported through two independent throttles** (TraceSegmentServiceClient.ts:174) — `logDiscardedBatch` at warn and `logReportError` at error, each with its own 30s window, so the same event appears twice at two levels and the two lines de-synchronise as the windows drift. Collapse to one throttled call carrying both the reason and the discarded count.
12. **`combineHostPort`'s doc overclaims parity** (BackendAddressResolver.ts:30): grpc-js's `uri-parser` version returns the bare host when port is undefined; this copy always appends `:${port}`. Since it is exported and unit-tested as a general helper, a caller passing the port-less `HostPort` that `splitHostPort` returns for a bare hostname gets `oap:undefined`. Narrow the doc to the bracketing rule, or mirror the missing branch.
13. **`npm ci --ignore-scripts` has an implicit allowlist** (tests/plugins/common/Dockerfile.agent:25): only grpc-tools is rebuilt, so the install scripts of every other dependency with one (currently `protobufjs`, `ssh2`) are skipped. Benign today, but the next dependency or transitive bump needing a native build will produce an image that fails at runtime inside the e2e jobs rather than at build time. Suppress scripts for the known-problematic package only, or assert the rebuilt set explicitly.

songzhendong:
Thanks for the Round-6 review.


### Fix

**Design**
1. Removed `format.splat()`; converted this PR’s `%s` log sites to template literals.
2. Scoped `retryPolicy` to unary `skywalking.v3.ManagementService/reportInstanceProperties` only (Management.proto FQDN; not Compat). Trace/meter `collect` and `keepAlive` are excluded.
3. Bounded flush waits at **2s** (`FLUSH_WAIT_MS`): Trace/Meter in-flight + report, and `agent.flush()` also bounds `SpanContext.flush()` so a hung RPC (or the flush request’s own open span) cannot stall hosts / Lambda for the full gRPC deadline.

**Fixes**
4. Restored the early `return` after `reportInstanceProperties` (mutually exclusive with `keepAlive`, same as master / Java).
5. MeterSender: once-only failure reporting + `stream.cancel()` (aligned with Trace).
6. Throttled connectivity transition logs.
7. `collectSample()` no longer overwrites a snapshot preserved across a disconnect window.
8. README documents that multi-address `sw-static` is literal endpoints only (no DNS expand / re-resolve; weaker than single `dns:`).
9. One-time boot warning when `SW_AGENT_SECURE` and multiple hostnames (authority pinned to the first).
10. E2E `/flush` deadline returns **504** + `flush deadline`; helpers assert `200` + `flushed`.
11. Collapsed discard logging to one throttled warn (reason + count).
12. `combineHostPort` mirrors grpc-js when port is missing (bare host / bracketed IPv6).
13. Dockerfile: after `npm ci --ignore-scripts`, explicitly `npm rebuild grpc-tools protobufjs ssh2` then `prepare`.

wu-sheng:
All three round-6 decisions landed correctly, verified: `splat()` is reverted with a comment recording why, every `%s` this PR added is now a template literal, `retryPolicy` is scoped to exactly `skywalking.v3.ManagementService/reportInstanceProperties`, the flush wait is bounded via a shared `FLUSH_WAIT_MS`, and the properties/keepAlive early return is restored with a note. No new design questions this round — most of what follows is my own round-6 wording being imprecise, corrected below.

**Design decisions:**

1. **Default the logging level to `warn`, and add a troubleshooting doc.** `warn` is the right level for discard/buffer-full diagnostics — but `createLogger` currently defaults to `error` and replaces `logger.warn` with a no-op below that level (logging/index.ts:74), so in a default deployment trace batches are discarded with no output at all, while MeterSender's `error`-level throttle keeps meter failures visible. Rather than raising those call sites, change the default level itself to `warn`, align MeterSender's report-failure throttle to `warn` so the two senders behave identically, and correct the README, which still documents the default as `info`.

   Some diagnostics not appearing at the default level is expected and fine — that just needs to be written down. Please add a troubleshooting doc under `docs/en/setup/` (mirroring the layout the other agents use, so it does not have to move later — see the note below) and link it from the README, covering:
   - which diagnostics appear at which level, and that lower-level lines are silent by design;
   - connectivity/failover lines, plus `GRPC_TRACE=pick_first,subchannel` for per-address detail;
   - what "buffer reached maximum size" and "discarded N segments" mean, and `SW_AGENT_MAX_BUFFER_SIZE`;
   - multi-backend + TLS: all backends must present certificates sharing SANs (verification follows the channel authority), with one DNS name → multiple A/AAAA records preferred for HA;
   - agent→OAP via HTTP proxy is unsupported and `grpc.enable_http_proxy=0` is set on every channel — this is an upgrade-affecting behavior change and currently lives inside the `SW_AGENT_LOGGING_LEVEL` README row, where nobody configuring a proxy will find it;
   - the best-effort contract: failures discard and are never re-sent, and `flush()` is a bounded attempt, not a delivery guarantee. This one explains every other symptom in the doc.

   (Separately, and explicitly **not** for this PR — I'll open an issue: this README is 189 lines and holds setup, the config table, runtime metrics, the supported-library list, Azure Functions, AWS Lambda and Webpack all at once, where `skywalking-python`'s is 54 and `skywalking-go`'s is 22 because they split content into `docs/en/` with a `docs/menu.yml` catalog. Correspondingly, `data/docs.yml` on the website links the Node.js agent straight to a GitHub tree — the only active agent without versioned, rendered docs under `content/docs/`. Restructuring the docs and wiring the website is its own piece of work and should not hold up this PR; putting the new page under `docs/en/setup/` now just means it lands where the restructure will want it.)

2. **After the bounded wait times out, start a genuinely new report.** `flush()`'s second step calls `reportOnce()`, which hits `coalesceReport`'s single-flight short-circuit and returns the *same* in-flight promise — so when the first wait times out (exactly the case the bound exists for), the segments buffered during the flush are never written to any stream, and flush burns a second timeout on a promise it already awaited. In Lambda the invocation's own segments are lost, while the doc comment above claims "one attempt at whatever is currently buffered"; the unit test passes only because it completes the first report manually. Bypass the coalescer on that path and issue a fresh `doReport`. `MeterSender.flush()` needs the same change for `latestSnapshot`.

3. **Give each connectivity transition its own throttle instance.** All three messages share one `logConnectivityTransition`, and `throttled()` keys solely on a timestamp — it never inspects the message. So the normal failover sequence (disconnect at t=0, recovered at t≈3s) logs the alarm and silently swallows the all-clear for 30s; if another disconnect follows after the window, the recovery line is lost outright and its count folded into the next disconnect's `suppressed` field. My round-6 ask to "wrap it in `throttled()` like its siblings" was wrong here: throttling a repeating error and throttling state transitions are different problems — transitions are paired events and each one matters. Separate instances for disconnect and recovered (or a message-keyed throttle) keep flap protection while guaranteeing the pair always appears.

**Fixes:**

4. **`coalesceReport`'s single-flight guard is not re-entrancy-safe** (coalesceReport.ts:62). `doReport()` is invoked before `state.reporting = flight` is assigned, and `doReport` runs synchronously through `emitter.emit('segments-sent')` whose listeners execute inline. A listener that reaches back into `flush()`/`reportOnce()` — both reachable from user code via `agent.flush()` — re-enters while the guard is still unset, starting a second `doReport()` that splices the buffer out from under the first; the outer assignment then orphans it so its `finally` never clears. Assign the promise before invoking `doReport`, or wrap the call in `Promise.resolve().then(...)`.

5. **The MeterSender sample guard is dead and distorts post-outage CPU** (MeterSender.ts:95). The retained snapshot is always overwritten on the first post-reconnect tick — the guard is false once CONNECTED and the timer samples before reporting — so it is never transmitted. Meanwhile skipping `collectSample()` freezes `RuntimeSampler`'s CPU delta baseline for the whole outage: after a 10-minute disconnect the first sample averages CPU over 600s instead of the 20s period, reporting a spike at roughly 1/30th magnitude. That is wrong data rather than missing data. Either drop the guard and keep sampling (the baseline stays correct), or make the report path actually send the preserved snapshot.

6. **Snapshot consumed before the config guard** (MeterSender.ts:128). `latestSnapshot = undefined` runs before the `!config.serviceName || !config.serviceInstance` early return — the same consume-before-check shape this PR deliberately fixed for the connectivity check one branch earlier. An embedder without a resolved `serviceInstance` samples every 20s, discards, and reports nothing, silently. Hoist the config guard above the consumption.

7. **`openChannel()` reassigns without closing the previous channel** (GRPCChannelManager.ts:246); only `shutdown()` ever calls `shutdownNow()`. A second `boot()` — an embedder restarting the agent, or any future re-open path, which this method's shape invites — leaks the old `grpc.Channel`: its sockets stay up, its subchannel pool reconnects with backoff forever, and its connectivity watcher keeps firing, silently returning on the identity check instead of unregistering. Close the existing channel before reassigning.

8. **The stream send/guard block is duplicated between the two senders** (TraceSegmentServiceClient.ts:176 and MeterSender.ts:135) — same collect → write → end → cancel → resolve shape with the same one-shot failure flag. The level split in item 1 is exactly this drift in action, and every future fix to the stream error path has to be applied twice. `coalesceReport.ts` was created as the shared home for this plumbing but only the single-flight guard was extracted; move the stream block there too.

9. **The E2E ambiguous case still degrades into a timeout** (helpers.ts:93). The JSDoc says "neither → throw for retry; both → fail loudly (ambiguous)", but both branches `throw` and the sole caller wraps the function in `waitForExpect` — so if the agent ever reports the same probe to both backends, which is precisely the bug this test exists to catch, it retries for the full 120s and reports a message indistinguishable from "collectors never came up". Let the ambiguous branch escape the retry loop: set a flag the caller asserts outside `waitForExpect`, or fail outside the retried block.

10. **The Dockerfile rebuild allowlist is hand-maintained and its comments contradict each other** (tests/plugins/common/Dockerfile.agent:31). `npm ci --ignore-scripts` is followed by `npm rebuild grpc-tools protobufjs ssh2`; a dependency added later with an install script is left half-installed and fails only when a plugin e2e container crashes on a missing native binding, pointing at the plugin rather than the image. `ssh2` is only there because `testcontainers` — a host-side devDependency never executed inside the container — pulls it in, so rebuilding it adds a native-build failure mode for no benefit. The rationale comment also appears twice (lines 22-23 and 27-28) with two different explanations, neither matching the three-package command. Scope script suppression to the package that actually breaks (or use `--omit=optional`), and collapse the comments to one accurate note.

11. **`getChannel()` on the manager is dead API** (GRPCChannelManager.ts:106). Every consumer now goes through `createClient()`; the watcher uses `getChannel()` on the inner `GRPCChannel`, not this method. Delete it along with its unreachable throw — leaving it invites a future caller to bypass `createClient`'s channelOverride contract.

songzhendong:
Thanks for the Round-7 review.


### Fix

**Design**
1. Default `SW_AGENT_LOGGING_LEVEL` is now `warn`; Meter report failures use a `warn` throttle (aligned with Trace discard). README default fixed; added `docs/en/setup/troubleshooting.md` (levels, connectivity/`GRPC_TRACE`, buffer/discard, TLS multi-backend, HTTP proxy, best-effort `flush`) and linked it from the README. No README overhaul / website docs in this PR.
2. After the bounded flush wait, Trace/Meter call `forceReport()` to start a new flight (bypasses single-flight coalesce) so data buffered during the wait is still attempted once.
3. Split connectivity throttles: `logChannelDisconnected` vs `logChannelRecovered`, so a recover line is not swallowed by the disconnect window.

**Fixes**
4. `coalesceReport` / `beginFlight`: assign `state.reporting` before invoking `doReport` (via `Promise.resolve().then`) so sync re-entry (e.g. `segments-sent` → flush) sees the guard.
5. Removed Meter “don’t sample while disconnected” — it never preserved a snapshot through reconnect (timer always re-samples) and froze CPU baseline across outages; always sample, only skip send when disconnected.
6. Meter: hoist `serviceName` / `serviceInstance` guard above snapshot consume so a misconfig does not silently drop the sample.
7. `openChannel`: `shutdownNow()` the previous channel after a successful rebuild (no leaked channel/watcher on re-entry).
8. Extracted shared client-streaming collect helper `runCollectStream` (once-only failure + `cancel`) used by Trace and Meter.
9. E2E: ambiguous probe (both collectors) returns `kind: 'ambiguous'` and fails outside `waitForExpect` instead of retrying for ~120s.
10. Dockerfile: drop unused `ssh2` rebuild; one accurate `--ignore-scripts` / rebuild comment; rebuild `grpc-tools protobufjs` only.
11. Removed dead `GRPCChannelManager.getChannel()`, plus dead `isConnected()`, `ServiceManagementClient.flush()`, and unused `_isInfoEnabled`.

wu-sheng:
*(Edited — the original version of this comment was written against a tree that predates #145 and asked for a docs layout that no longer applies. Please follow the version below.)*

**Please rebase onto the latest `master` and adopt the docs structure that is already there.**

This branch is based on `6522a26` (#143), which predates #145 ("docs: move user documentation into docs/ for website hosting"). `master` now carries the full layout — `docs/menu.yml` plus `docs/en/{setup,features,plugins,advanced,contribution}/` — so the pages this PR needs already exist and only need extending:

1. **Do not add `docs/en/setup/troubleshooting.md`.** `master` already has **`docs/en/advanced/troubleshooting.md`**, and `docs/menu.yml` points "Troubleshooting" at `/en/advanced/troubleshooting`. A second page would be invisible on the website and would split the same topic in two. Fold this PR's content into the existing page — its current sections ("OAP connection errors", "Traces are missing during an OAP outage", "Agent logs are hard to find", "Data is missing when the process stops") already cover the ground this change affects, so most of it is extending those rather than adding new headings. What needs to land there: which diagnostics appear at which log level (and that lower-level lines are silent by design); the disconnect/recovered lines and `GRPC_TRACE=pick_first,subchannel` for per-address detail; what "buffer reached maximum size" and "discarded N segments" mean, with `SW_AGENT_MAX_BUFFER_SIZE`; multi-backend under TLS requiring certificates that share SANs (verification follows the channel authority, so prefer one DNS name with multiple A/AAAA records); agent→OAP over an HTTP proxy being unsupported with `grpc.enable_http_proxy=0` on every channel — an upgrade-affecting change that warrants its own heading; and the best-effort contract (failures discard and are never re-sent, `flush()` is a bounded attempt rather than a delivery guarantee), which explains every symptom above it.

2. **The environment-variable table now lives in `docs/en/setup/configuration.md`, not the README.** Two rows there need this PR's changes: `SW_AGENT_COLLECTOR_BACKEND_SERVICES` still reads "OAP gRPC address in `host:port` form" and needs the comma-separated multi-backend semantics (single address → grpc-js `dns:`, multiple → `pick_first` over a static resolver) plus the TLS caveat; and `SW_AGENT_LOGGING_LEVEL` still documents the default as `error`, which this PR changes to `warn`.

3. **No `docs/menu.yml` change is needed** — every page this PR touches is already listed.

4. **Drop the `#`-prefixed ASF header from the new markdown.** None of the pages `master` added carry one, and in Markdown those lines render as a stack of `<h1>`s on the website.

Everything else from my other comment (the code items) stands unchanged.

songzhendong:
Thanks for the follow-up on the docs layout after

Rebased onto latest `master` and moved the documentation to the existing pages (no second troubleshooting file, no `menu.yml` change).

### Fix

1. Removed `docs/en/setup/troubleshooting.md`. Folded this PR’s content into **`docs/en/advanced/troubleshooting.md`**: log-level visibility (default `warn`, lower levels silent by design); disconnect/recovered + `GRPC_TRACE=pick_first,subchannel`; buffer full / discarded segments + `SW_AGENT_MAX_BUFFER_SIZE`; multi-backend under TLS (authority / shared SANs); a dedicated **Agent to OAP over an HTTP proxy** section (`grpc.enable_http_proxy=0`); and the best-effort / bounded `flush()` contract in the existing stop/flush guidance.
2. Updated **`docs/en/setup/configuration.md`**: `SW_AGENT_COLLECTOR_BACKEND_SERVICES` documents single-address `dns:` vs comma-separated static `pick_first` + TLS caveat; `SW_AGENT_LOGGING_LEVEL` default is `warn`. Removed the outdated “uses only the first entry” note.
3. Left `docs/menu.yml` unchanged.
4. No `#`-prefixed ASF header on the markdown pages (matches the #145 docs style).
5. README stays on the #145 structure (env table lives in `configuration.md`).

wu-sheng:
The rebase onto `master` landed correctly — the branch now sits on #145, there is a single `docs/en/advanced/troubleshooting.md` with no duplicate page, and all eleven round-7 code items remain fixed. Type-check is clean and the unit suite passes 47/47. I also verified the transport behaviour empirically this round: a real channel built with this PR's `sw-static` resolver and service config fails over to the standby and self-heals out of sticky `TRANSIENT_FAILURE` against live and killed local gRPC servers.

Remaining items, most of them in the flush plumbing added since the last review.

**Correctness:**

1. **The two flush budgets are not coherent** (src/index.ts:69). `agent.flush()` caps the whole service flush at `FLUSH_WAIT_MS`, but `TraceSegmentServiceClient.flush()` can spend `FLUSH_WAIT_MS` waiting on the in-flight report and *then* start the forced report with another `FLUSH_WAIT_MS` — so the inner path may consume twice the constant the outer path allows. The outer wait expires exactly when the forced attempt begins.

   To be clear about how narrow the real impact is: with a healthy backend everything completes in milliseconds and the cap never engages, and with a uniformly slow backend a fresh report would stall too, so the data is lost either way. The case where it changes the outcome is failover: the in-flight call is stuck on the dead backend until its 10s deadline while the channel has already reconnected to the standby, so a *new* report would go out on the healthy connection and succeed — which is precisely what `forceReport` was added for. Today it starts too late to be awaited, and in Lambda the process freezes right after `agent.flush()` resolves.

   Worth fixing mainly because the fix is trivial and the current shape is self-inconsistent: derive the outer bound from what the inner path can consume, or have both share a single deadline. Note this is not Lambda-specific — `agent.flush()` is public API and shutdown hooks hit the same path; the Lambda plugin is just the only automatic caller.

2. **A timed-out SpanContext flush leaks its resolver** (src/index.ts:65). `SpanContext.flush()` pushes `resolve` onto the static `SpanContext.flushResolve` array, and that array is drained only when `nTotalSegments` reaches 0. Bounding the wait with `awaitWithTimeout` abandons the promise but never removes the entry, so a long-lived server that calls `agent.flush()` on a request path while concurrent requests keep `nTotalSegments > 0` accumulates one resolver plus its Promise per call, growing until traffic happens to drain to zero. Master awaited unconditionally, so the entry was always eventually consumed. Have the timeout path remove its own resolver, or give `SpanContext.flush()` its own timeout contract.

3. **An empty catch swallows `segments-sent` listener exceptions** (TraceSegmentServiceClient.ts:155). The new try/catch discards listener errors with no log, counter or debug line. The plugin e2e suites rely on this event, so a throwing listener surfaces later as a missing-assertion timeout with nothing in the log pointing at it. A `logger.debug` or throttled warn in the catch preserves the intent — don't reject the report promise — without hiding the fault.

**Logging:**

4. **A successful reconnect is logged at ERROR level** (GRPCChannelManager.ts:37). `logChannelRecovered` uses `throttled(logger, 'error', 30000)`, so the healthy end of a failover raises an error-severity line. Splitting the throttle buckets was the right fix from round 7; the level is not — alerting keyed on `level: error` (the natural rule, since the agent's only other error lines are auth rejection and disconnect) fires on every recovery. `info` or `warn` carries the same information.

5. **The discard log counts calls, not segments** (TraceSegmentServiceClient.ts:187). `throttled`'s `suppressed` counter increments per log call, and each emitted line names only that batch's length. During a 5-minute outage the 1s loop can discard up to `maxBufferSize` segments per attempt, but the operator sees ~10 lines reading "Discarded 3 trace segment(s)" with `suppressed: 29` and no way to learn that tens of thousands were dropped. Since best-effort discarding is now the design (restore was deliberately removed), this line is the only signal that it is happening — please carry a monotonic dropped-segment total in the throttled message so the loss is measurable.

6. **`ServiceManager` still uses `%s`** (lines 54, 62, 70, 82). With `format.splat()` deliberately absent, winston treats the Error as meta and its message/stack are non-enumerable, so the record is `{"message":"ServiceManager boot failed: %s"}` with no detail. This predates the PR, but it matters more now that `GRPCChannelManager.boot()` can throw and the default level moved to `warn` — and the PR converted its own `%s` sites while leaving these four. Convert them to template literals for consistency.

**Optional simplification** (fine as a follow-up if you would rather not churn this PR further): grpc-js's pick_first supports `shuffleAddressList` in the LB config this PR already emits, so `{ pick_first: { shuffleAddressList: true } }` would replace the hand-rolled `shuffleAddresses` helper, its export and its dedicated test — **and** remove the `config.secure ? [...parsed] : shuffleAddresses(parsed)` special case, its warning log and the accompanying docs, because the TLS-authority problem only exists since the shuffle is applied to the target string. Shuffling at the LB layer leaves the target — and therefore the authority and SNI — stable. It would also shuffle the multi-A/AAAA endpoint list of the single-address `dns:` target, which the current helper cannot reach.

**Cleanups:**

7. `MeterSender.flush()` and `TraceSegmentServiceClient.flush()` are the same wait-then-forceReport body with one predicate changed — the budget fix in item 1 has to be applied twice and can drift, which is how the warn/error split appeared two rounds ago. Extract a `flushCoalesced(state, doReport, isClosed, hasPending)` next to `coalesceReport`/`forceReport`.
8. The outer try/catch in `doReportBufferedMetrics` (MeterSender.ts:166) is unreachable: the only statements inside it are field reads and `runCollectStream(...)`, which handles `open`/`writeAll`/`end` internally and never rethrows. Remove it, or have `runCollectStream` rethrow and keep one handler.
9. `previous?.shutdownNow()` in `openChannel` (GRPCChannelManager.ts:238) cannot execute — the only call site is `boot()`, guarded by `ServiceManager.booted`, so `managedChannel` is always null there. Drop it, or keep it with a comment saying it is defensive against a future re-open path.
10. `combineHostPort`'s port-less branch (BackendAddressResolver.ts:33) is unreachable: `tryParseHostPort`, its only production caller, already rejects `port == null` and out-of-range ports. Inlining the bracket-and-join into `tryParseHostPort` removes an exported symbol, a dead branch and a test.
11. The `../../../src:/app/src` bind mount (tests/remote-e2e/static-failover/docker-compose.yml:51) shadows the `src/proto` stubs the image's `npm run prepare` generated, so the container silently depends on the host checkout having run protoc — and if it has not (missing `protocol/` submodule, grpc-tools download failure), the only symptom is the `server-1` health check timing out with no useful message. Mount only the sources the test edits, or exclude `src/proto`.
12. The `.catch(() => undefined)` at MeterSender.ts:95 and TraceSegmentServiceClient.ts:132 are redundant — `coalesceReport` documents and implements "never rejects". Drop them, or drop the contract from the helper's doc and keep the guards deliberately; as-is the redundant guard will keep multiplying at each new call site.

songzhendong:
Thanks for the follow-up review.
Addressed the flush plumbing / logging / cleanup items, plus the optional `shuffleAddressList` simplification.
### Fix
**Correctness**
1. Introduced `flushCoalesced()` with one shared `FLUSH_WAIT_MS` deadline for wait-in-flight + `forceReport` (Trace and Meter). `agent.flush()` still bounds SpanContext and the service flush at `FLUSH_WAIT_MS` each, so the outer wait is never shorter than what the inner path can consume.
2. `SpanContext.flush(timeoutMs)` removes its own resolver on timeout (and on success) so abandoned flush waiters cannot accumulate.
3. `segments-sent` listener failures are logged at `debug` without rejecting the report promise.
**Logging**
4. Channel recovered uses a `warn` throttle (disconnect stays `error`) so recovery does not trip error-level alerts.
5. Discard lines include a monotonic `total discarded` count across the process lifetime.
6. Converted the four `ServiceManager` `%s` sites to template literals.
**Optional**
- Replaced hand-rolled `shuffleAddresses` with `pick_first: { shuffleAddressList: true }`. The target string keeps config order (stable TLS authority / SNI); endpoint pick order is shuffled by grpc-js. Removed the insecure/TLS shuffle branch, the boot warning that existed only because the target was shuffled, and the dedicated shuffle unit test. Docs updated accordingly.
**Cleanups**
7. Trace/Meter flush both call `flushCoalesced`.
8. Removed the unreachable outer try/catch in Meter `doReportBufferedMetrics`.
9. Kept `previous?.shutdownNow()` in `openChannel` with a comment that boot is once today but re-open must not leak channels.
10. Inlined host:port formatting into `tryParseHostPort`; dropped exported `combineHostPort` and the dead port-less branch.
11. Remote-e2e no longer bind-mounts host `src/` over the image (preserves `npm run prepare`’s `src/proto`).
12. Dropped redundant `.catch(() => undefined)` on coalesceReport callers (helper never rejects).

wu-sheng:
Clean round — eleven of the twelve items are fixed, and the optional simplification was taken as well. Verified individually: `flushCoalesced` now shares one deadline across both phases and `SpanContext.flush()` takes its own timeout; the abandoned resolver removes itself from the static array under a `settled` guard with an unref'd timer; `segments-sent` listener errors are logged rather than swallowed; recovery moved from `error` to `warn`; discards carry a monotonic `discardedSegmentTotal`; `ServiceManager`'s four `%s` sites are gone; the unreachable try/catch, `combineHostPort` and the redundant `.catch`es are deleted; and the compose file no longer bind-mounts over the generated protos, with a comment recording why. Adopting `pick_first: { shuffleAddressList: true }` removed the hand-rolled shuffle helper *and* the whole `config.secure` special case — nice result. Keeping `previous?.shutdownNow()` as defensive code is fine.

One item left, and it is the same narrow failover window as before rather than anything new in kind.

**The shared budget can leave the forced report with zero time** (coalesceReport.ts:128). Sharing one deadline is the right fix and it does remove the 2× overrun. But when the in-flight wait consumes the whole budget — which is exactly the stuck-call case this path exists for — `remaining()` returns 0, so phase 2 calls `forceReport(...)` and then `awaitWithTimeout(..., 0)` resolves on the next timer tick. The forced report is genuinely started, but `flush()` returns without awaiting it, and in Lambda the process freezes mid-flight. So in the failover scenario the net outcome is unchanged from before the fix: the in-flight call is stuck on the dead backend, phase 1 burns the full budget waiting on it, and the fresh report that would have succeeded on the already-reconnected standby never gets awaited.

Two things make this straightforward to resolve:

- The priority is inverted relative to the semantics we agreed on. The in-flight batch has **already had its one attempt**; the pending buffer has not. Under "ensure one attempt, not guaranteed delivery", the un-attempted data has the stronger claim on the budget.
- Phase 2 does not depend on phase 1 at all — `forceReport` bypasses the single-flight guard by calling `beginFlight` directly, so it can start a new report while the old one is still outstanding. Phase 1's wait is a courtesy to the already-attempted batch, not a precondition.

So either reserve a slice for phase 2 (cap the in-flight wait at, say, half the budget), or run phase 2 first and give the in-flight wait whatever remains. Either shape keeps the total bounded by one `FLUSH_WAIT_MS` while guaranteeing the un-attempted data gets a real, awaited attempt.

Nothing else outstanding from my side.

songzhendong:
Thanks for the follow-up on the remaining `flushCoalesced` budget issue.
The shared-budget path that could leave `forceReport` with zero await time is fixed.
### Fix
- In `flushCoalesced`, force pending first (`forceReport`), then wait on the prior in-flight promise with whatever budget remains.
- Pending has priority: the in-flight batch already had one attempt; total wait stays within one `FLUSH_WAIT_MS`.

wu-sheng (approved):
LGTM

